PR #4, opened 2023-03-29
Super simple demo script to make things easy

kartiknagappa:
Adding a super-simple-low-tech demo script which makes it easy for users to run a prediction using the Sync CLI. The low-tech aspect makes it easy to see the Sync CLI commands used to run a prediction, query status, and get results.

Adding logic to loop on status etc added unnecessary complexity and diminished the Sync CLI.
Keeping it super simple made the most sense.


malino (approved):

rmoneys:
To make this easier to use add a she-bang at the top and make it executable:
```
#!/bin/bash
...
```
```
% chmod u+x demo_emr.sh
```

PR #7, opened 2023-03-29
Resolve correct Python

rmoneys:
Default to `python` if `python3` doesn't point to the right version in the installation script.


kartiknagappa (approved):
LGTM

PR #8, opened 2023-03-31
Add asyncemr module

rmoneys:
Renames async api package and starts an asyncemr module to mirror the corresponding synchronous module.


CaymanWilliams (approved):
LGTM

PR #12, opened 2023-04-12
[PROD-806] Update sync_sdk to fetch & upload new cluster report data for Databricks alongside eventlog submissions

gseyffert-sync:
Fetch new Databricks cluster report information to submit alongside event logs

Make event log uploading more reliable

Add Databricks AWS Region static configuration


sync/awsdatabricks.py
```diff
@@ -99,11 +138,37 @@ def create_prediction_for_run(
...
+
+        cluster_events = get_default_client().get_cluster_events(cluster_id)
+
+        aws_region_name = get_databricks_config().aws_region_name
+        ec2 = boto.client("ec2", region_name=aws_region_name)
+        instances = ec2.describe_instances(
+            Filters=[
+                {"Name": "tag:Vendor", "Values": ["Databricks"]},
+                {"Name": "tag:ClusterId", "Values": [cluster_id]},
+                # {'Name': 'tag:JobId', 'Values': []}
```
gseyffert-sync, line 156: I was considering adding this in the call, but it's not strictly necessary. Looking at it in the light of day, I'm also not 100% sure if this would play nicely with shared compute clusters vs. single-job ones. Perhaps best to leave it out

sync/awsdatabricks.py
```diff
@@ -525,6 +590,7 @@ def wait_for_final_run_status(run_id: str) -> Response[str]:
 
         sleep(30)
         run = get_default_client().get_run(run_id)
+        # TODO - log status periodically?
```
gseyffert-sync, line 593: This comment is me wondering if it makes sense to periodically log the state of the Databricks Job while we're waiting. This doesn't really add anything beyond giving a user of the CLI some confirmation that things are still progressing.

sync/cli/__init__.py
```diff
@@ -52,25 +52,35 @@ def configure():
...
+        dbx_token = click.prompt(
+            "Databricks token",
+            default=DB_CONFIG.token if DB_CONFIG else OPTIONAL_DEFAULT,
+            hide_input=True,
+            show_default=False,
+        )
+        dbx_region = click.prompt(
+            "Databricks AWS Region name",
+            default=DB_CONFIG.aws_region_name if DB_CONFIG else OPTIONAL_DEFAULT,
+        )
```
gseyffert-sync, line 72: Instead of requiring users to specify a `region_name` as an argument to the various Databricks functions, I thought it would be a bit nicer if they could just configure that alongside their DBX host/token. Since a Databricks workspace can only exist in one region, this value is pretty static.

sync/clients/databricks.py
```diff
@@ -94,12 +143,22 @@ def _send(self, request: httpx.Request) -> dict:
...
 
-_sync_client: DatabricksClient = None
+_databricks_config: DatabricksConf | None = None
+
+
+def get_databricks_config() -> DatabricksConf:
+    global _databricks_config
+    if not _databricks_config:
+        _databricks_config = DatabricksConf()
+    return _databricks_config
```
gseyffert-sync, line 153: I split this out so that I could query it independently here - https://github.com/synccomputingcode/syncsparkpy/pull/12/files#diff-fcb826fa7c84469280d545c2fb63c0ad6e4d83d38a3cff66ac7cd3d4c6b0ea5eR150

sync/cli/awsdatabricks.py
```diff
@@ -63,9 +63,7 @@ def run_job(job_id: str, plan: str, compute: str, project: dict = None):
 @click.option("--project", callback=validate_project)
 def create_prediction(run_id: str, plan: str, compute: str, project: str = None):
     """Create a prediction for a job run"""
-    prediction_response = awsdatabricks.create_prediction_for_run(
-        run_id, plan, compute, project["id"]
```
gseyffert-sync, line 67: little bug here trying to index a string

sync/awsdatabricks.py
```diff
@@ -92,7 +92,7 @@ def create_cluster(config: dict) -> Response[str]:
     :return: cluster ID
     :rtype: Response[str]
     """
-    response = get_default_client().create_cluster(config)
+    response = dbx.get_default_client().create_cluster(config)
```
gseyffert-sync, line 95: Made this change to faciliate `@patch`ing in tests

sync/awsdatabricks.py
```diff
@@ -706,6 +710,11 @@ def _s3_contents_have_all_rollover_logs(contents: list[dict], run_end_time_secon
     )
 
 
+def event_log_poll_duration_seconds():
+    """Convenience function to aid testing"""
+    return 15
```
gseyffert-sync, line 715: Also for testing, so that I can easily set this to `0` to avoid needing to wait forever when testing the polling logic

tests/test_awsdatabricks.py
```diff
@@ -0,0 +1,597 @@
...
+
+import boto3 as boto
+from botocore.response import StreamingBody
+from botocore.stub import Stubber
+
+from sync.awsdatabricks import create_prediction_for_run
+from sync.config import DatabricksConf
+from sync.models import Response
+
+MOCK_RUN = {
```
gseyffert-sync, line 15: I've always preferred defining mock data inline, if the size is not completely unwieldy

tests/test_awsdatabricks.py
```diff
@@ -0,0 +1,597 @@
...
+
+    assert not result.error
+    assert result.result
+
+
+@patch("sync.clients.databricks.get_databricks_config")
+@patch("sync.clients.databricks.get_default_client")
+@patch("sync.awsdatabricks.event_log_poll_duration_seconds")
+@patch("sync.awsdatabricks.create_prediction")
+def test_create_prediction_for_run_event_log_upload_delay(
```
gseyffert-sync, line 484: This test is intended to exercise the various scenarios in which we want to continue polling the S3 event log location. These conditions include -

- No event log data found at all yet,
- Incomplete event log data where the `eventlog` file (with no suffix) is missing,
- Incomplete event log data where the `LastModified` time of the event log is prior to the Run finishing

sync/__init__.py
```diff
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.0.2"
+__version__ = "0.0.3"
```
rmoneys, line 2: I think we'll want to force-push v0.0.3 - jumped the gun on pushing that up. Probably ok as long as there aren't any big interface changes.

sync/awsdatabricks.py
```diff
@@ -4,30 +4,62 @@
...
+        Instances should be narrowed to just those instances relevant to the Databricks Run. This can be done by passing
+        a `tag:ClusterId` filter to the describe_instances call like -
+        Filters=[
+            {"Name": "tag:ClusterId", "Values": ["my-dbx-clusterid"]}
+        ]
+
+        If there are multiple pages of instances, all pages should be accumulated into 1 dictionary and passed to this
+        function
+    :type instances: dict
+
```
rmoneys, line 62: Getting an unexpected indentation warning when trying to build the docs that seems to come from this. Also, I'm not sure I'm a fan of the blank lines between parameters in the interest of screen space, inconsistency with autogenerated docstrings and presentation is more of a concern for sphinx.

rmoneys (commented):
I'm not sure it's a good use of time to go through a draft PR from top to bottom. I cut this review short after that realization. Let me know what areas I should look at while this is a draft. I'll go through it from top to bottom once you think it's ready to merge.

sync/__init__.py
rmoneys: nm. let's make this 0.0.4. I think at least Pete is using v0.0.3 right now... don't want to pull the rug out from anyone...

gseyffert-sync:
Notable changes in this PR include -

- The addition of a new `get_cluster_report` CLI action - https://github.com/synccomputingcode/syncsparkpy/pull/12/files#diff-65ab4a0ddbcebe2782bef80db856e5a7ff183439b6eb159dcc332fa0a362197aR84
- Updates to what data is fetched & included as part of the Databricks Cluster Report. Reasoning for the new data can be found here - https://synccomputing.atlassian.net/wiki/spaces/EN/pages/806780933/Input+Spark+Application+Cost+Estimation+Cluster+Timelines#Databricks
- Improved resiliency for event log upload lag in `awsdatabricks._get_eventlog` - https://github.com/synccomputingcode/syncsparkpy/pull/12/files#diff-fcb826fa7c84469280d545c2fb63c0ad6e4d83d38a3cff66ac7cd3d4c6b0ea5eR731
- The addition of an `aws_region_name` value to `DatabricksConf`. This is necessary because calls like `ec2.describe_instances` require specifying a `region_name`, but we can't reliably get this value from e.g. the cluster definition. However, DBX workspaces exist only within 1 AWS Region, so statically configuring this should work well enough without requiring the user to pass it to every function call,
- A `get_cluster_events` addition to `clients/databricks`,

sync/awsdatabricks.py
rmoneys: I think `@patch("sync.awsdatabricks.get_default_client")` would have worked.

It is nice to have that additional context telling us it's a databricks client, but that can be had by aliasing get_default_client in the import statement.

rmoneys: Also, I started mocking a bit lower down: https://github.com/synccomputingcode/syncsparkpy/blob/main/tests/api/test_predictions.py

It gives us more coverage, and doesn't require a mock client.

rmoneys (commented):
I didn't get very far in this review, but I think it's worth considering a testing approach...

sync/clients/databricks.py
```diff
@@ -48,6 +48,55 @@ def delete_cluster(self, cluster_id: str) -> dict:
             )
         )
 
+    def get_cluster_events(
+        self, cluster_id: str, start_time_ms: int | None = None, end_time_ms: int | None = None
+    ):
```
rmoneys, line 53: This kinda sets a new precedent for this module. The idea is that client methods are 1 to 1 with API transactions. That said, I think this is the first we needed to page through API responses... an alternative is a private function in awsdatabricks 🤷

I'm ok with it

sync/clients/databricks.py
rmoneys: I'm not sure this really belongs here. I've been doing,
```
from sync.config import CONFIG
```
which is provided by: https://github.com/synccomputingcode/syncsparkpy/blob/2919c60058acd73889f767d97593ec87dec62d69/sync/config.py#L148-L172

sync/awsdatabricks.py
```diff
@@ -92,21 +121,74 @@ def create_prediction_for_run(
...
+                )
 
-    return cluster_response
+            return eventlog_response
+        return cluster_report_response
+    return cluster_id_response
+
+
+def get_cluster_report(cluster_id: str, plan_type: str, compute_type: str) -> Response[dict]:
+    """ """
```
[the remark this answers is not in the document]
gseyffert-sync: Ah, you are correct.

sync/awsemr.py
```diff
@@ -378,17 +378,17 @@ def create_prediction_for_cluster(cluster_id: str, region_name: str = None) -> R
     return cluster_report
 
 
-def get_cluster_report(cluster_id: str, region: str = None) -> Response[dict]:
+def get_cluster_report(cluster_id: str, region_name: str = None) -> Response[dict]:
```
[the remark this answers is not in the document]
gseyffert-sync: The other functions in this module also use `region_name` by my eye

sync/cli/awsdatabricks.py
gseyffert-sync: Ah - the type hint on the argument was `str`, so I was going off of that. Sounds like that type hint should be a `dict` instead, then?

sync/clients/databricks.py
gseyffert-sync: ah, duh - so I should be able to just `from sync.config import DB_CONFIG`?

sync/cli/awsdatabricks.py
gseyffert-sync: I see - I missed that `validate_project` callback. Makes more sense now! I was like "how is a user reasonably going to be able to pass a `dict` here?"

sync/awsdatabricks.py
```diff
@@ -623,16 +703,71 @@ def _get_task_cluster(task: dict, clusters: list) -> Response[dict]:
...
-            Bucket=parsed_log_url.netloc,
-            Prefix=f"{parsed_log_url.path.strip('/')}/{task.get('cluster_instance', {}).get('cluster_id')}/eventlog/{task.get('cluster_instance').get('cluster_id')}",
-        ).get("Contents")
+
+        if stripped_path := parsed_log_url.path.strip("/"):
+            prefix = f"{stripped_path}/{cluster_id}/eventlog/{cluster_id}"
+        else:
+            # If the event log destination is just a *bucket* without any sub-path, then we don't want to include
+            #  a leading `/` in our Prefix (which will make it so that we never actually find the event log)
+            prefix = f"{cluster_id}/eventlog/{cluster_id}"
```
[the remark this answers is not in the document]
gseyffert-sync: I don't think that will work, because if `stripped_path == ""`, then we are pre-prending a `/` to our `Prefix`, which is what causes the bug. If the log path is _just_ a bucket, then we don't want any leading `/` in there at all. I love getting rid of `if` statements, but I don't see a great way to for this one...

sync/awsdatabricks.py
gseyffert-sync: > I think @patch("sync.awsdatabricks.get_default_client") would have worked.

I did try this first, and it didn't work, and so after poking around a little bit I found this - https://stackoverflow.com/questions/20242862/why-python-mock-patch-doesnt-work. It's interesting that it seems to work for the tests in `test_awsemr`, though...

sync/awsdatabricks.py
gseyffert-sync: > I think it better to check for contents before sleeping in _get_eventlog. ...  or maybe you can put the check in the loop statement....

I agree! And yeah, we have the check in the loop - we skip the `sleep` on `num_attempts == 0`, so the first call is made immediately. If we are satisfied with the contents in the bucket, then we just move on without ever calling `sleep`!

I still need to keep this here, though, because I have a test case that _is_ testing the polling scenario, and I don't want that test to take like a minute to run.


> In my limited manual testing the event log was immediately available

So, there is a little bit of complexity here. For longer-running jobs, there will always be SOME eventlog data available. The only question is if ALL of the eventlog data has been flushed. Databricks should flush all the data upon cluster termination, but sometimes we actually catch the cluster while it's still RUNNING and not yet TERMINATED. I linked this in a few comments in code, but they have some documentation here about this - https://docs.databricks.com/clusters/configure.html#cluster-log-delivery-1

So without any polling mechanism, we can fool ourselves into thinking that we have everything we need, when in reality we are missing a bit of the event log data. I actually caught this because I was testing using a very _short_ job (about ~2 minutes to execute), and was frequently getting a "No eventlog found at location" error (even though it _was_ there when I looked manually)

sync/awsdatabricks.py
```diff
@@ -623,16 +703,71 @@ def _get_task_cluster(task: dict, clusters: list) -> Response[dict]:
...
+    """Convenience function to aid testing"""
+    return 15
+
+
+def _get_eventlog(task: dict, cluster: dict, run_end_time_millis: int) -> Response[bytes]:
     log_url = cluster.get("cluster_log_conf", {}).get("s3", {}).get("destination")
     if log_url:
         parsed_log_url = urlparse(log_url)
-
+        cluster_id = task.get("cluster_instance", {}).get("cluster_id")
```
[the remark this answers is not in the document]
gseyffert-sync: True - or just pass in the known cluster ID that we check for in `create_prediction_for_run`. We already do the cluster ID error handling there, and we don't use the `task` argument for anything else in this function, by my eye. Since this is a private function, I think requiring a non-`None` argument is reasonable.

sync/cli/awsdatabricks.py
```diff
@@ -63,10 +64,31 @@ def run_job(job_id: str, plan: str, compute: str, project: dict = None):
...
+@click.argument("cluster-id")
+@click.option(
+    "--plan", type=click.Choice(["Standard", "Premium", "Enterprise"]), default="Standard"
+)
+@click.option(
+    "--compute",
+    type=click.Choice(["All-Purpose Compute", "Jobs Compute", "Jobs Light Compute"]),
+    default="Jobs Compute",
+)
+def get_cluster_report(cluster_id: str, plan: str, compute: str):
```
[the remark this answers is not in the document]
gseyffert-sync: Perhaps, though I suppose it's kind of a semantic argument - since this is `get_cluster_report`, I think passing the `cluster_id` matches the expected behaviour the closest. However, yeah, we could make this `get_run_cluster_report` instead, and have it accept a `run_id`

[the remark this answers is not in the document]
rmoneys: We don't necessarily have to change the command name (get-run-cluster-report is kinda long). I'm just thinking, I know how to get the run ID easily in the Databricks UI, but I don't know how to get the cluster ID off the top of my head. We could put it out and solicit feedback from Pete, but my hunch is get-cluster-report that takes a run ID is gonna be easier to digest. My 2 cents - not asking for a change.

sync/awsdatabricks.py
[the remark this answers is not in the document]
rmoneys: That's the reason for the strip at the end of the statement:
```
>>> url = "s3://megacorp-jobs"
>>> cluster_id = "0411-011041-ldaw3bkd"
>>> stripped_path = urlparse(url).path
>>> f"{stripped_path}/{cluster_id}/eventlog/{cluster_id}".strip("/")
'0411-011041-ldaw3bkd/eventlog/0411-011041-ldaw3bkd'
>>> url = "s3://megacorp-jobs/projects"
>>> stripped_path = urlparse(url).path
>>> f"{stripped_path}/{cluster_id}/eventlog/{cluster_id}".strip("/")
'projects/0411-011041-ldaw3bkd/eventlog/0411-011041-ldaw3bkd'
```

sync/awsemr.py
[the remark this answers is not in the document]
rmoneys: Yeah, I tried to consistently align on region_name, but didn't quite get there

sync/cli/awsdatabricks.py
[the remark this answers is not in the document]
gseyffert-sync: No, it's a good point - the Run ID is definitely easier to get from the Databricks UI vs. a cluster ID.

sync/clients/databricks.py
gseyffert-sync: Hmm - this does bring up the question of how to easily alter this global constant for tests. I guess some sort of monkeypatching?

sync/awsdatabricks.py
gseyffert-sync: I think I misunderstood your comment when I first read it. So the patch should target the imported name _in the module we're testing_, not in the `clients.databricks` module. That makes more sense

gseyffert-sync: And indeed, that does work 🎉  thanks for the suggestion!

sync/cli/__init__.py
```diff
@@ -35,9 +35,9 @@ def main(debug: bool):
 @main.command
 def configure():
     """Configure Sync Library"""
-    api_key_id = click.prompt("API key ID", default=API_KEY.id if API_KEY else None)
+    api_key_id = click.prompt("Sync API Key ID", default=API_KEY.id if API_KEY else None)
```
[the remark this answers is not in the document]
gseyffert-sync: Yeah, good point - those should probably be updated to be consistent. I guess I don't have a _strong_ opinion one way or the other... however, I feel like most CLIs I've used have been `Title Cased` when configuring them? I'm not sure. I think it reads a little more polished, personally.

Ironically, the Databricks CLI seems to use a combination of both casing conventions -

Screenshot 2023-04-13 at 12 32 05 PM

sync/clients/databricks.py
rmoneys: Yeah, I was thinking something similar to what we do in sync_backend, e.g. https://github.com/synccomputingcode/sync_backend/blob/ca26316173212ec225cefef803c463d00cc5d721/tests/unit/spark_predictor/web_tools/test_aws_ebs_pricing.py#L58-L62

sync/cli/awsdatabricks.py
rmoneys: Yeah, looks like I got the type descriptor wrong

sync/awsdatabricks.py
rmoneys: I do wait on the cluster when running a one-off job (with either project or prediction configuration applied): https://github.com/synccomputingcode/syncsparkpy/blob/2919c60058acd73889f767d97593ec87dec62d69/sync/awsdatabricks.py#L557

One alternative that may be worth considering is adding a wait on the cluster in addition to the job run when we're just executing an existing job (/2.1/jobs/run-now).

gseyffert-sync: yeah, and I was testing with `run-job`, which should be calling into this. I'm glad you pointed this out, though, because I'd missed it - it seems to me that `existing_cluster_id` only exists if you are running a Job/Task on a cluster that was already provisioned. If instead a new cluster is launched when you start a Job run, `existing_cluster_id` will not be present -

existing_cluster_id | string If  existing_cluster_id, the ID of an existing cluster that is used for all  runs of this task. When running tasks on an existing cluster, you may  need to manually restart the cluster if it stops responding. We suggest  running jobs on new clusters for greater reliability.
-- | --

https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsRunNow

I don't necessarily want to block on the cluster terminating _here_, because we do want to make the EC2 `describe_instances` call as soon as possible, but that could be a good alternative to polling S3, for sure

gseyffert-sync: Even still, we would likely still want to keep the polling for the `LastModified` time, though. Since we know now that users can configure log destinations in regions other than the one their DBX workspace is in, that additional cross-region latency will likely cause consistency issues for us in some cases

sync/cli/__init__.py
```diff
@@ -52,25 +52,35 @@ def configure():
...
-        hide_input=True,
-        show_default=False,
-    )
+    dbx_host = OPTIONAL_DEFAULT
+    dbx_token = OPTIONAL_DEFAULT
+    dbx_region = OPTIONAL_DEFAULT
+    if click.confirm("Would you like to configure a Databricks workspace?"):
+        dbx_host = click.prompt(
+            "Databricks host (prefix with https://)",
+            default=DB_CONFIG.host if DB_CONFIG else OPTIONAL_DEFAULT,
```
rmoneys, line 61: In playing around with this branch I tried running an existing command,
```
% sync-cli aws-databricks run-job --project load_all_lending_tables 926350091914688
```
which errored out because I didn't have a valid Databricks configuration:
```
% cat ~/.sync/databrickscfg
{
  "host": "https://dbc-d95d06ca-1d00.cloud.databricks.com",
  "token": "derpdedoooo"
}
```
The required region is missing. So I ran `sync-cli configure` which seemed to work, but didn't populate defaults from the existing configuration because it's invalid:
```
% sync-cli configure
Sync API Key ID [SXfC5pbTuHB8FPPy5psUHPgTg7cS9SPm]: 
Sync API Key Secret: 
Default project S3 URL [s3://sync-us-east-1-temp-90-days/scott/projects]: 
Default prediction preference (performance, balanced, economy) [balanced]: 
Would you like to configure a Databricks workspace? [y/N]: y
Databricks host (prefix with https://) [none]:
```
I'd be nice to version these config files to allow for graceful upgrades (I kinda want to get ride of the optional `default_project_url` too). That may not be the right use of resources now though so maybe Pete can weigh in on whether this experience is ok.

gseyffert-sync: yeah, I had a similar experience when testing this, and I agree, that would be the best experience! I think the tricky part of this is how to make this new field _required_, but still allow loading the old config that may be missing this newly required field. To allow Pydantic to load it, we'd need to make the model field optional, but the downside is that it's hard to kind of force the user to re-configure to specify this new required data.

rmoneys: Maybe we could just default to the AWS region:
```
aws_region_name: str = Field(boto.client("s3").meta.region_name, env="DATABRICKS_AWS_REGION")
```

gseyffert-sync: That is an interesting thought 🤔 then we can display the current default in that prompt? I'll give it a try! Good idea

sync/cli/__init__.py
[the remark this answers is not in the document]
kartiknagappa: 👉  Sentence case.

tl;dr;
I quickly looked at a few examples (AWS CLI, Heroku CLI). Appears to be some general inconsistencies. Heroku uses sentence case and lowercase. AWS has some inconsistency between the UI and CLI.

For example, in AWS
- UI // `Access key` `Secret access key`
- `aws configure` prompt // `AWS Access Key ID` `AWS Secret Access Key`
- AWS CLI credentials file // `aws_access_key_id` `aws_secret_access_key`

sync/awsdatabricks.py
rmoneys: Yeah, just ran a quick test. Looks like for runs based on an existing job (/2.1/jobs/run-now) the cluster ID will be at .cluster_instance.cluster_id for tasks that are at the PENDING or later lifecycle stage...

sync/cli/__init__.py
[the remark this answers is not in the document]
gseyffert-sync: Sentence case it is, then!

sync/awsdatabricks.py
```diff
@@ -76,6 +110,7 @@ def get_cluster(cluster_id: str) -> Response[dict]:
     return Response(result=cluster)
 
 
+# TODO - Databricks configuration documentation
 def create_prediction_for_run(
     run_id: str, plan_type: str, compute_type: str, project_id: str = None
```
gseyffert-sync, line 115: I defined the `plan_type` and `compute_type` enums, but haven't updated these functions to use them everywhere, yet... this diff is already getting a little large

tests/test_awsdatabricks.py
```diff
@@ -0,0 +1,625 @@
...
+    host="https://dbc-123.cloud.databricks.com",
+    token="my_secret_token",
+    aws_region_name="us-east-1",
+)
+
+
+def test_create_prediction_for_run_failed_run(respx_mock):
+    failure_response = {"error_code": "FAILED", "message": "This run failed"}
+
+    respx_mock.get("https://*.cloud.databricks.com/api/2.1/jobs/runs/get?run_id=75778").mock(
```
gseyffert-sync, line 314: @rmoneys updated all these tests to use `respx` to mock out our HTTP calls instead :+1:

gseyffert-sync:
Getting some 429s in the Test action - 

```
   /usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +4b829b55a0ac77cc7ee1a6175484d9ed289b3303:refs/remotes/pull/12/merge
  Error: fatal: unable to access 'https://github.com/synccomputingcode/syncsparkpy/': The requested URL returned error: 429
  The process '/usr/bin/git' failed with exit code 128
  Waiting 20 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +4b829b55a0ac77cc7ee1a6175484d9ed289b3303:refs/remotes/pull/12/merge
  Error: fatal: unable to access 'https://github.com/synccomputingcode/syncsparkpy/': The requested URL returned error: 429
  The process '/usr/bin/git' failed with exit code 128
  Waiting 10 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +4b829b55a0ac77cc7ee1a6175484d9ed289b3303:refs/remotes/pull/12/merge
  Error: fatal: unable to access 'https://github.com/synccomputingcode/syncsparkpy/': The requested URL returned error: 429
  Error: The process '/usr/bin/git' failed with exit code 128
```

rmoneys:
> Getting some 429s in the Test action

Damn maybe we need to distribute this to pypi sooner than I thought.

Re-request a review from me when you think this is ready to merge.

tests/test_awsdatabricks.py
rmoneys: Love it!

sync/clients/databricks.py
```diff
@@ -48,6 +48,21 @@ def delete_cluster(self, cluster_id: str) -> dict:
             )
         )
 
+    def get_cluster_events(self, cluster_id: str, **kwargs) -> dict:
```
gseyffert-sync, line 51: Using `**kwargs` here to allow passing arbitrary query params for e.g. paging requests. I didn't want to be too "strict" since Databricks can always decide to send back any random query params for the `next_page` that they'd like to

gseyffert-sync (commented):
@rmoneys this should be ready for another pass whenever you're ready!

rmoneys (approved):

sync/clients/databricks.py
rmoneys: 👍 "params" might be more descriptive than "kwargs". I was thinking the optional params might be better encapsulated in a dict like elsewhere in this module, but the POST params make up all the method args so not so important.

- just my 2 cents, no expectation about further changes to this PR

PR #15, opened 2023-04-24
Adjusted for cluster definition in task payload.

gorskysd:
Also publicized `get_cluster_report` method for external access.

Also tweaked timing on validation for getting cluster report (see comment).

### DIFF
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -251,7 +251,7 @@ def get_prediction_job(
         job = get_default_client().get_job(job_id)
         if "error_code" in job:
             return Response(error=DatabricksAPIError(**job))
-        
+
         job_settings = job["settings"]
         if tasks := job_settings.get("tasks", []):
             cluster_response = _get_job_cluster(tasks, job_settings.get("job_clusters", []))
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -182,8 +182,7 @@ def get_cluster_report(
 
     # Making these calls prior to fetching the event log allows Databricks a little extra time to finish
     #  uploading all the event log data before we start checking for it
-    cluster_events = get_default_client().get_cluster_events(cluster_id)
-
+    cluster_events = _get_all_cluster_events(cluster_id)
     aws_region_name = DB_CONFIG.aws_region_name
     ec2 = boto.client("ec2", region_name=aws_region_name)
 
@@ -743,7 +742,7 @@ def _s3_contents_have_all_rollover_logs(contents: list[dict], run_end_time_secon
         # can proceed with the upload + prediction. If this proves to be finicky in some way, we can always
         # wait for the full 5 minutes after the run_end_time_seconds
         final_rollover_log
-        and final_rollover_log["LastModified"].timestamp() >= (run_end_time_seconds - 15)
+        and final_rollover_log["LastModified"].timestamp() >= run_end_time_seconds
     )
 
 
@@ -812,7 +811,7 @@ def _get_eventlog(cluster_description: dict, run_end_time_millis: int) -> Respon
         )
 
 
-def get_all_cluster_events(cluster_id: str):
+def _get_all_cluster_events(cluster_id: str):
     """Fetches all ClusterEvents for a given Databricks cluster, optionally within a time window.
     Pages will be followed and returned as 1 object
     """
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -252,7 +252,6 @@ def get_prediction_job(
         job = get_default_client().get_job(job_id)
         if "error_code" in job:
             return Response(error=DatabricksAPIError(**job))
-
         job_settings = job["settings"]
         if tasks := job_settings.get("tasks", []):
             cluster_response = _get_job_cluster(tasks, job_settings.get("job_clusters", []))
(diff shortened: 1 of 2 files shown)


sync/awsdatabricks.py
```diff
@@ -739,7 +743,7 @@ def _s3_contents_have_all_rollover_logs(contents: list[dict], run_end_time_secon
         # can proceed with the upload + prediction. If this proves to be finicky in some way, we can always
         # wait for the full 5 minutes after the run_end_time_seconds
         final_rollover_log
-        and final_rollover_log["LastModified"].timestamp() >= run_end_time_seconds
+        and final_rollover_log["LastModified"].timestamp() >= (run_end_time_seconds - 15)
```
gorskysd, line 746: @gseyffert-sync Curious if you think this is ok. In the case of the jobs I tested on the last eventlog modification was consistently ~10 sec before the run end time.

sync/awsdatabricks.py
```diff
@@ -808,7 +812,7 @@ def _get_eventlog(cluster_description: dict, run_end_time_millis: int) -> Respon
         )
 
 
-def _get_all_cluster_events(cluster_id: str):
+def get_all_cluster_events(cluster_id: str):
```
gorskysd, line 815: Made this public so I could save the full cluster events.

@gseyffert-sync I'm curious if it was a deliberate decision to not use this method in the `get_cluster_report` method. It seems the current state of that method will not return the full cluster if it is over a certain number of events.

sync/awsdatabricks.py
```diff
@@ -252,20 +252,21 @@ def get_prediction_job(
         job = get_default_client().get_job(job_id)
         if "error_code" in job:
             return Response(error=DatabricksAPIError(**job))
-
         job_settings = job["settings"]
         if tasks := job_settings.get("tasks", []):
             cluster_response = _get_job_cluster(tasks, job_settings.get("job_clusters", []))
             if cluster := cluster_response.result:
                 prediction_cluster = _deep_update(
                     cluster, prediction["solutions"][preference]["configuration"]
```
gorskysd, line 260: I noticed a sneaky inconsistency here between the SDK and the Prediction API.

The `get_project_job` method sets the log url to `s3://root/path/{{project_id}}` via the `get_project_cluster_settings` method.  However, Predictions API return a new payload with the log url reset to just `s3://root/path` (no project_id appended to the path). It's worth making this consistent, and IMO having the API also return the s3_url with the `project_id` included makes a lot of sense. Curious if anyone else has thoughts here.

This isn't blocking any job bank work.

sync/awsdatabricks.py
gseyffert-sync: ah, this was actually just a miss - the intent is definitely to use this method (originally I had this logic in the `awsdatabricks` client method, but didn't update that call site when I added this function).

sync/awsdatabricks.py
gseyffert-sync: was the prior condition causing you to not be able to upload event logs? In general, I'd prefer to code against documented behaviour (as in the link in this comment) vs. observed behaviour

sync/awsdatabricks.py
gseyffert-sync: If we just fix `get_cluster_report`, does this function still need to be public?

gorskysd: Nope, I'll make that change and make this method private again.

sync/awsdatabricks.py
gseyffert-sync: do Job Bank runs not terminated the cluster at the end of the run? Because event logs are also purged when the cluster is terminated. In my manual tests, I didn't need to wait 5 minutes - only need to poll once or twice after the job finished (so, 15 - 30 seconds)

rmoneys: Let's at least log when we fail to meet this condition before exhausting attempts. A warning sounds appropriate.

gseyffert-sync: We do log each time we poll (below on line 780), but it's an `info` log right now.

sync/awsdatabricks.py
gseyffert-sync: I'm on board with that, but if we append the project ID in the API response, won't the URL that is visible to users be different than the one they configured anyway? Or do you mean that only `GET /predictions/{id}` should append the project ID, but `GET /projects/{id}` can return the "raw" path?

rmoneys: > Or do you mean that only GET /predictions/{id} should append the project ID, but GET /projects/{id} can return the "raw" path?

Yes

sync/awsdatabricks.py
gorskysd: Just reverted this change.

Job Bank doesn't explicitly terminate the cluster because it uses the `new_cluster` field of the JobCreate payload. The cluster terminates after the run, it just so happens in the case I looked at that the eventlog is last updated in s3 a little bit before the cluster terminates.

sync/awsdatabricks.py
```diff
@@ -252,20 +251,21 @@ def get_prediction_job(
         job = get_default_client().get_job(job_id)
         if "error_code" in job:
             return Response(error=DatabricksAPIError(**job))
-
```
rmoneys, line 255: Is this intentional? I thought the line break helped set apart the error handling/early exit from the meat of the function.

rmoneys (dismissed):

rmoneys:
https://synccomputing.atlassian.net/browse/PROD-856

gseyffert-sync (approved):

sync/awsdatabricks.py
gorskysd: Think this was accidental when I was mucking about. Reverted.

rmoneys (approved):

PR #16, opened 2023-04-26
Fix some errors that occur when applying cluster recommendations to DBX jobs

gseyffert-sync:
When attempting to apply one of our recommendations to a pre-existing Databricks job, I hit a few errors, so this is just a little PR to resolve those

### DIFF
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -389,6 +389,8 @@ def run_job_object(job: dict) -> Response[str]:
 
     if cluster := cluster_response.result:
         if len(tasks) == 1:
+            # For `new_cluster` definitions, Databricks will automatically assign the newly created cluster a name,
+            #  and will reject any run submissions where the `cluster_name` is pre-populated
             if "cluster_name" in cluster:
                 del cluster["cluster_name"]
 
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -283,9 +283,6 @@ def get_prediction_job(
 
                 prediction_cluster = _deep_update(cluster, prediction["solutions"][preference]["configuration"])
 
-                if "cluster_name" in prediction_cluster:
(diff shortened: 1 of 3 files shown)


sync/awsdatabricks.py
```diff
@@ -271,9 +271,21 @@ def get_prediction_job(
...
-                )
+                # num_workers/autoscale are mutually exclusive settings, and we are relying on our Prediction
+                #  Recommendations to set these appropriately. Since we may recommend a Static cluster (i.e. a cluster
+                #  with `num_workers`) for a cluster that was originally autoscaled, we want to make sure to remove this
+                #  prior configuration
+                if "num_workers" in cluster:
+                    del cluster["num_workers"]
+
+                if "autoscale" in cluster:
+                    del cluster["autoscale"]
```
gseyffert-sync, line 282: Remove any old cluster sizing configuration so that we aren't providing e.g. _both_ `num_workers` AND `autoscale`

sync/awsdatabricks.py
```diff
@@ -271,9 +271,21 @@ def get_prediction_job(
...
+
+                prediction_cluster = _deep_update(cluster, prediction["solutions"][preference]["configuration"])
+
+                if "cluster_name" in prediction_cluster:
+                    del prediction_cluster["cluster_name"]
```
gseyffert-sync, line 287: Was running in to this error from the DBX API - 

> Databricks API Error: INVALID_PARAMETER_VALUE: Cluster name should not be provided for jobs.

gseyffert-sync: I think this is safe to blanketly delete, since we are _always_ submitting Jobs here...

rmoneys (dismissed):
Looks good, but I think a comment about why we're removing cluster name would be helpful. Something like, "'cluster_name' isn't allowed in embedded cluster definitions"

rmoneys (approved):

PR #17, opened 2023-05-03
PROD-952 pip3 address in projects page setup needs to change

gseyffert-sync:
Giving this Action a try to auto-publish a `latest` version at the same time we publish updated documentation - https://github.com/EndBug/latest-tag

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Publish Documentation
+name: Publish Documentation & Latest Tag
 
 # Controls when the workflow will run
 on:
@@ -31,6 +31,8 @@ jobs:
       - uses: actions/upload-pages-artifact@v1
         with:
           path: docs/_build/html
+      - name: Run latest-tag
+        uses: EndBug/latest-tag@latest
 
   deploy:
     # Add a dependency to the build job


gseyffert-sync:
I'm not sure how to test this prior to merging this PR 🤔

gseyffert-sync:
An alternative we could use instead, which works for tags other than just `latest` - https://github.com/marketplace/actions/actions-tagger

rmoneys (commented):
That's a cool workflow step: https://github.com/EndBug/latest-tag. I added it to the ticket: https://synccomputing.atlassian.net/browse/PROD-910

This doesn't guarantee that latest points to a release. Users might use it to download a revision between releases (tags) which could make it difficult to troubleshoot issues, e.g. `sync-cli --version` reports "0.0.2" but really they're running something from a couple commits after that. How about we manually set it until the above ticket is resolved (hopefully next sprint).

gseyffert-sync:
yeah - I'm wondering if I should just look at `actions-tagger` and potentially just address the whole release "action" in this PR 🤔 if you don't mind me picking up that ticket

rmoneys:
> yeah - I'm wondering if I should just look at actions-tagger and potentially just address the whole release "action" in this PR 🤔 if you don't mind me picking up that ticket

Please do!

PR #18, opened 2023-05-04
PROD-958 Update handling of incomplete cluster_report data for Databricks jobs

gseyffert-sync:
First pass at allowing creation of cluster reports that may be missing instance data due to e.g. the run being too old. This adds the `--allow-incomplete` boolean flag on both `get-cluster-report` and `create-prediction` CLI commands & their associated library functions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
   "httpx~=0.23.0",
   "orjson~=3.8.0",
   "click~=8.1.0",
+  "tenacity==8.2.2"
 ]
 dynamic = ["version", "description"]
 
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -221,7 +221,8 @@ def _get_cluster_report(
     if not instances["Reservations"]:
         no_instances_message = (
             f"Unable to find any active or recently terminated instances for cluster `{cluster_id}` in `{aws_region_name}`. "
-            + "Please refer to the following documentation for options on how to address this - <some link>"
+            + "Please refer to the following documentation for options on how to address this - "
+            + "https://synccomputingcode.github.io/syncsparkpy/reference/awsdatabricks.html"
         )
         if allow_incomplete:
             logger.warning(no_instances_message)
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -112,7 +112,11 @@ def get_cluster(cluster_id: str) -> Response[dict]:
 
 # TODO - Databricks configuration documentation
 def create_prediction_for_run(
-    run_id: str, plan_type: str, compute_type: str, project_id: str = None
+    run_id: str,
+    plan_type: str,
+    compute_type: str,
+    project_id: str = None,
+    allow_incomplete_cluster_report: bool = False,
 ) -> Response[str]:
     """Create a prediction for the specified Databricks run.
 
@@ -124,6 +128,8 @@ def create_prediction_for_run(
     :type compute_type: str
     :param project_id: Sync project ID, defaults to None
     :type project_id: str, optional
+    :param allow_incomplete_cluster_report: Whether creating a prediction with incomplete cluster report data should be allowable
+    :type allow_incomplete_cluster_report: bool, optional, defaults to False
     :return: prediction ID
     :rtype: Response[str]
     """
@@ -138,7 +144,9 @@ def create_prediction_for_run(
     if cluster_id := cluster_id_response.result:
         # Making these calls prior to fetching the event log allows Databricks a little extra time to finish
         #  uploading all the event log data before we start checking for it
-        cluster_report_response = _get_cluster_report(cluster_id, plan_type, compute_type)
+        cluster_report_response = _get_cluster_report(
+            cluster_id, plan_type, compute_type, allow_incomplete_cluster_report
+        )
         if cluster_report := cluster_report_response.result:
 
             cluster = cluster_report.cluster
@@ -161,7 +169,7 @@ def create_prediction_for_run(
 
 
 def get_cluster_report(
-    run_id: str, plan_type: str, compute_type: str
+    run_id: str, plan_type: str, compute_type: str, allow_incomplete: bool = False
 ) -> Response[DatabricksClusterReport]:
     """Fetches the cluster information required to create a Sync prediction
 
@@ -171,6 +179,8 @@ def get_cluster_report(
     :type plan_type: str
     :param compute_type: Cluster compute type, e.g. "Jobs Compute"
     :type compute_type: str
+    :param allow_incomplete: Whether creating a cluster report with incomplete data should be allowable
+    :type allow_incomplete: bool, optional, defaults to False
     :return: cluster report
     :rtype: Response[DatabricksClusterReport]
     """
@@ -183,13 +193,13 @@ def get_cluster_report(
 
     cluster_id_response = _get_run_cluster_id(run["tasks"])
     if cluster_id := cluster_id_response.result:
-        return _get_cluster_report(cluster_id, plan_type, compute_type)
+        return _get_cluster_report(cluster_id, plan_type, compute_type, allow_incomplete)
 
     return cluster_id_response
 
 
 def _get_cluster_report(
-    cluster_id: str, plan_type: str, compute_type: str
+    cluster_id: str, plan_type: str, compute_type: str, allow_incomplete: bool
 ) -> Response[DatabricksClusterReport]:
     cluster = get_default_client().get_cluster(cluster_id)
     if "error_code" in cluster:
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
   "httpx~=0.23.0",
   "orjson~=3.8.0",
   "click~=8.1.0",
+  "tenacity==8.2.2"
```
gseyffert-sync, line 34: Small library I'm pulling in for easily adding retry logic beyond what `httpx` allows. This is used down here - https://github.com/synccomputingcode/syncsparkpy/pull/18/files#diff-a92790f88cf634b038c61957033a487fc111749ec671d4821d0ac692fb69a94fR114

(`httpx` actually recommends this library in their docs as well - https://www.python-httpx.org/advanced/#usage_1)

sync/awsdatabricks.py
```diff
@@ -112,7 +112,11 @@ def get_cluster(cluster_id: str) -> Response[dict]:
 
 # TODO - Databricks configuration documentation
 def create_prediction_for_run(
-    run_id: str, plan_type: str, compute_type: str, project_id: str = None
+    run_id: str,
+    plan_type: str,
+    compute_type: str,
+    project_id: str = None,
+    allow_incomplete_cluster_report: bool = False,
```
gseyffert-sync, line 119: named the parameter a little differently here to be a bit more descriptive

sync/awsdatabricks.py
```diff
@@ -209,11 +219,14 @@ def _get_cluster_report(
         ]
     )
     if not instances["Reservations"]:
-        return Response(
-            error=DatabricksError(
-                message=f"Unable to find any active or recently terminated instances for cluster `{cluster_id}` in `{aws_region_name}`"
-            )
+        no_instances_message = (
+            f"Unable to find any active or recently terminated instances for cluster `{cluster_id}` in `{aws_region_name}`. "
+            + "Please refer to the following documentation for options on how to address this - <some link>"
```
gseyffert-sync, line 224: This message will need to be updated to point to the correct documentation page. For now I've left a placeholder in.

sync/awsdatabricks.py
```diff
@@ -281,7 +294,9 @@ def get_prediction_job(
                 if "autoscale" in cluster:
                     del cluster["autoscale"]
 
-                prediction_cluster = _deep_update(cluster, prediction["solutions"][preference]["configuration"])
+                prediction_cluster = _deep_update(
+                    cluster, prediction["solutions"][preference]["configuration"]
+                )
```
gseyffert-sync, line 299: auto-formatting

sync/cli/__init__.py
```diff
@@ -23,7 +23,7 @@ def main(debug: bool):
     if debug:
         logging.basicConfig(level=logging.INFO, format=LOG_FORMAT)
     else:
-        logging.basicConfig(level=logging.CRITICAL, format=LOG_FORMAT)
+        logging.basicConfig(level=logging.WARNING, format=LOG_FORMAT)
```
gseyffert-sync, line 26: log level change, as discussed

sync/clients/sync.py
```diff
@@ -109,7 +110,16 @@ def delete_project(self, project_id: str) -> dict:
         return self._send(self._client.build_request("DELETE", f"/v1/projects/{project_id}"))
 
     def _send(self, request: httpx.Request) -> dict:
-        response = self._client.send(request)
+        try:
+            for attempt in Retrying(stop=stop_after_attempt(3), wait=wait_fixed(5)):
```
gseyffert-sync, line 114: Adding some retry logic for `503`s so that API calls to e.g. dev/stage don't fail if our instances are sleeping.

sync/clients/sync.py
```diff
@@ -109,7 +110,16 @@ def delete_project(self, project_id: str) -> dict:
...
+                with attempt:
+                    response = self._client.send(request)
+                    if response.status_code == httpx.codes.SERVICE_UNAVAILABLE:
+                        # Doesn't need to be any particular exception, this is just to trigger the retry
+                        raise Exception()
```
rmoneys, line 119: Looks like `tenacity.TryAgain` was made to be used in places like this.

sync/clients/sync.py
```diff
@@ -109,7 +110,16 @@ def delete_project(self, project_id: str) -> dict:
...
+        except RetryError:
+            # If we max out on retries, then just bail and log the error we got
+            pass
```
rmoneys, line 122: This will ignore any exceptions from the block won't it?

sync/clients/sync.py
rmoneys: This aspect of our dev environment is probably better handled with DNS configuration to park the request until the server's ready per the documentation, but retrying is good to have in general. Is this a good time to all the clients here? Additionally, expand the scope to include other status codes: {408, 425, 429, 500, 502, 503, 504} and an incrementing wait to recover from tiny blips faster.

rmoneys (dismissed):
Successfully tested, but I have some questions about retrying.

sync/clients/sync.py
gseyffert-sync: ah, nice find!

sync/clients/sync.py
gseyffert-sync: yeah, this was a little step towards adding some more general retrying for resiliency purposes. I agree with ya about there probably being a better way to solve this for dev/stage. I'd be happy to add additional status codes to this diff while we're at it

sync/clients/__init__.py
```diff
@@ -15,3 +17,54 @@ def encode_json(obj: dict) -> tuple[dict, str]:
         "Content-Length": str(len(json)),
         "Content-Type": "application/json",
     }, json
+
+
+class RetryableHTTPClient:
```
gseyffert-sync, line 22: Instead of copy/pasting the retry logic to the clients we have, I figured we could just have a little base class that our clients could inherit from as appropriate.

Alternatively, these could just be functions that accept a client and a request as input.

sync/clients/databricks.py
```diff
@@ -92,9 +96,9 @@ def get_run(self, run_id: str) -> dict:
         )
 
     def _send(self, request: httpx.Request) -> dict:
-        response = self._client.send(request)
+        response = self._send_request(request)
 
-        if response.status_code >= 200 and response.status_code < 300:
+        if 200 <= response.status_code < 300:
```
gseyffert-sync, line 101: PyCharm suggested this little change to me

sync/clients/__init__.py
```diff
@@ -15,3 +17,54 @@ def encode_json(obj: dict) -> tuple[dict, str]:
...
+    def _send_request(self, request: httpx.Request) -> httpx.Response:
+        try:
+            for attempt in Retrying(
+                stop=stop_after_attempt(3),
+                wait=wait_exponential_jitter(initial=2, max=10, jitter=2),
+            ):
+                with attempt:
+                    response = self._client.send(request)
+                    if response.status_code in self._DEFAULT_RETRYABLE_STATUS_CODES:
+                        raise TryAgain
```
gseyffert-sync, line 49: `TryAgain` seems to do the trick as well :+1:

sync/clients/databricks.py
rmoneys: Nice. Didn't know you could do that. Doesn't make a lot of sense to me from an operator associativity standpoint...

sync/clients/sync.py
rmoneys: I'm saying that without `reraise=True` exceptions that arise from within the `with attempt` block will be "wrapped" by a `RetryError` and ignored, and response won't be set. Not a big problem, but less than ideal IMO.

sync/clients/__init__.py
rmoneys: Do you think it's a typo in their documentation to not instantiate `TryAgain`? I don't think I've seen that before. Oh, maybe just not recommended: https://peps.python.org/pep-0317/

gseyffert-sync: Ah, yeah, I was curious about that... it did work when I tried, though!

sync/clients/sync.py
gseyffert-sync: ah, I see what you're saying! That's most likely true - let me give it a try to double check

sync/awsdatabricks.py
rmoneys: Probably better not to log TODOs (some link). It could be there for a while.

gseyffert-sync: Yeah - let me point to the CLI docs for now, and update this when we have some more comprehensive documentation that goes over some different use cases.

sync/clients/__init__.py
rmoneys: Works for me too. Interesting though that after the DNS change I get a 502 back from staging instead of a 503...

rmoneys (approved):
Manual testing looks good. Thanks for the updates!

sync/clients/databricks.py
rmoneys: https://docs.python.org/3/reference/expressions.html#comparisons

PR #19, opened 2023-05-09
PROD-910 Define and implement release process for syncsparkpy

gseyffert-sync:
This PR adds a new Action for releasing versions of `syncsparkpy`. A few things that I tried to achieve -

- Immutable releases
  - Once a release has been created, we don't want to "update" it. If we need to patch it, a new release should be created
- Singular version value
  - The `__version__` we set in the library should match the published release version
- No need to manually push tags
  - As above, manually pushing tags creates the opportunity to "fat finger" the tag

I'm sure we can add to this as we go, but I think this works for now. The `latest` and `v0.0.8` tags/releases were published while testing out this action, and can be seen here - https://github.com/synccomputingcode/syncsparkpy/tags

Note - I was able to test this action by putting this PR into Draft mode and configuring it with a `pull_request` trigger. Then pushes to my branch would automatically re-trigger the action, allowing me to observe its behaviour.


.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
+name: Release new library version
+
+# Controls when the workflow will run
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
gseyffert-sync, line 6: Leave this as a manual action for now

.github/workflows/pull-request.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python 3.10.x
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
```
gseyffert-sync, line 18: GitHub was giving me a warning about the old version using a deprecated version of NodeJS (version 12), so just bumped this version in order to resolve that issue

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Check release version doesn't yet exist
+        run: |
+          if [ $(git tag -l '${{ env.tag_version }}') ]; then echo "Release already found - aborting" && exit 1; else echo "Continuing with release"; fi
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ env.tag_version }}
+          generate_release_notes: true
+      - name: Run latest-tag
+        uses: EndBug/latest-tag@latest
```
gseyffert-sync, line 47: Summarizing this -

- Check out the full repo history (necessary for checking tags that may already exist),
- Install python / library deps,
- Create an env var named `tag_version` that is simply `v{__version__}`,
- Verify that we haven't already created a tag for this version
  -  If we have, the build will fail and the version will need to be incremented
- Create release `v{__version__}`
- Tag this release as `latest`

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+
+  build-docs:
+    # Only publish docs if we were able to successfully create a release
+    needs: release-latest
```
gseyffert-sync, line 51: Gate this so that we don't publish docs unless we could actually create a release

.github/workflows/docs.yml
```diff
@@ -1,54 +0,0 @@
-name: Publish Documentation
```
gseyffert-sync, line 1: Deleted this action since now it's part of the `release` action

PR #20, opened 2023-05-23
Allow tasks to be excluded from cluster reports

rmoneys:
Allow excluding tasks when getting a cluster report for a prediction. This allows workflow integration by tacking on a "Sync" task to record the run/create a prediction.

### DIFF
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -111,7 +111,6 @@ def get_cluster(cluster_id: str) -> Response[dict]:
     return Response(result=cluster)
 
 
-# TODO - Databricks configuration documentation
 def create_prediction_for_run(
     run_id: str,
     plan_type: str,
@@ -258,7 +257,14 @@ def _get_cluster_report(
     )
 
 
-def record_run(run_id: str, plan_type: str, compute_type: str, project_id: str) -> Response[str]:
+def record_run(
+    run_id: str,
+    plan_type: str,
+    compute_type: str,
+    project_id: str,
+    allow_incomplete_cluster_report: bool = False,
+    exclude_tasks: Collection[str] | None = None,
+) -> Response[str]:
     """See :py:func:`~create_prediction_for_run`
 
     :param run_id: Databricks run ID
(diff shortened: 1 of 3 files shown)


sync/awsdatabricks.py
```diff
@@ -181,17 +192,23 @@ def get_cluster_report(
...
-        return Response(error=DatabricksError(message="Run did not successfully complete"))
+    tasks = [
+        task for task in run["tasks"] if not exclude_tasks or task["task_key"] not in exclude_tasks
+    ]
 
-    cluster_id_response = _get_run_cluster_id(run["tasks"])
+    if any(task["state"].get("result_state") != "SUCCESS" for task in tasks):
+        return Response(error=DatabricksError(message="Tasks did not complete successfully"))
+
+    cluster_id_response = _get_run_cluster_id(tasks)
```
gseyffert-sync, line 211: Just to double-check my understanding here -

- This allows us to get "in-progress" runs because we are no longer looking at the `run` state itself, but instead the state of each of the tasks that are not in the exclusion list,
- Polling for the eventlog will be fine because we are expecting this task to run on some other compute resource, so the cluster the non-excluded tasks run on should terminate as expected
  - Should we enforce this uniqueness? This exclusion mechanism is more generic than a "special task" one, so requiring this may not quite fit if you want to exclude other tasks for any reason

gseyffert-sync (dismissed):
Optimistically approving this.

I like this approach! Gives us the ability to show users how to set up a "special task" for eventlog collection/submission at the end of their jobs, but doesn't require us to hard-code any special identifier in the library.

sync/awsdatabricks.py
gseyffert-sync: I'm just thinking though that if we _don't_ enforce this somehow, we could accidentally cause an "infinite wait", since the cluster that we are polling will be the same one that we're running on

gseyffert-sync (approved):
:+1: seems pretty reasonable! thanks for adding a state check/timeout

PR #22, opened 2023-05-26
Sync Agent `init_script`

gseyffert-sync:
This PR adds a `monitor_cluster` CLI command for Databricks that -

- Checks for a cluster log destination,
- If one exists, polls the `ec2.describes_instances` API every 30 seconds to discover instances attached to the current cluster,
  - For each successful poll, dedupes the current snapshot with our previous snapshot (see - https://github.com/synccomputingcode/syncsparkpy/compare/init_scripts#diff-fcb826fa7c84469280d545c2fb63c0ad6e4d83d38a3cff66ac7cd3d4c6b0ea5eR906-R926)
- Saves each snapshot to a file `/sync_data/cluster_instances.json` in the cluster log destination directory

Then, when grabbing a cluster report, we first check if this `/sync_data/cluster_instances.json` file exists. If it does, we will include that in the cluster report. Otherwise, we will still attempt to take a snapshot in order to get _some_ data.

For full context, I've published the associated `init_script` in a gist here - https://gist.github.com/gseyffert-sync/4b91525affe327e836fe9ac94f349b02

### DIFF
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -1,6 +1,7 @@
 """
 Utilities for interacting with Databricks
 """
+import datetime
 import io
 import logging
 import time
@@ -12,6 +13,7 @@ from typing import Any, TypeVar
 from urllib.parse import urlparse
 
 import boto3 as boto
+from orjson import orjson
 
 from sync.api.predictions import create_prediction_with_eventlog_bytes, get_prediction
 from sync.api.projects import get_project
@@ -336,10 +338,11 @@ def get_prediction_job(
 
                 if cluster_key := tasks[0].get("job_cluster_key"):
                     job_settings["job_clusters"] = [
-                        j
-                        for j in job_settings["job_clusters"]
-                        if j.get("job_cluster_key") != cluster_key
-                    ] + [{"job_cluster_key": cluster_key, "new_cluster": prediction_cluster}]
+                                                       j
+                                                       for j in job_settings["job_clusters"]
+                                                       if j.get("job_cluster_key") != cluster_key
+                                                   ] + [{"job_cluster_key": cluster_key,
+                                                         "new_cluster": prediction_cluster}]
                 else:
                     tasks[0]["new_cluster"] = prediction_cluster
                 return Response(result=job)
@@ -377,10 +380,11 @@ def get_project_job(job_id: str, project_id: str, region_name: str = None) -> Re
                 project_cluster = _deep_update(cluster, project_cluster_settings)
                 if cluster_key := tasks[0].get("job_cluster_key"):
                     job_settings["job_clusters"] = [
-                        j
-                        for j in job_settings["job_clusters"]
-                        if j.get("job_cluster_key") != cluster_key
-                    ] + [{"job_cluster_key": cluster_key, "new_cluster": project_cluster}]
+                                                       j
+                                                       for j in job_settings["job_clusters"]
+                                                       if j.get("job_cluster_key") != cluster_key
+                                                   ] + [
+                                                       {"job_cluster_key": cluster_key, "new_cluster": project_cluster}]
                 else:
                     tasks[0]["new_cluster"] = project_cluster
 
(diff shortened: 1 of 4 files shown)


sync/awsdatabricks.py
```diff
@@ -799,6 +852,89 @@ def _wait_for_cluster_termination(
...
+                    old_instances = [res for res in previous_instances["Reservations"]]
+                    old_instance_id_to_reservation = dict(
+                        zip(
+                            [res["Instances"][0]["InstanceId"] for res in old_instances],
+                            old_instances,
+                        )
+                    )
+
+                    old_instance_ids = set(old_instance_id_to_reservation.keys())
+                    new_instance_ids = set(new_instance_id_to_reservation.keys())
```
rmoneys, line 900: I don't think `.keys` is needed

sync/awsdatabricks.py
```diff
@@ -799,6 +852,89 @@ def _wait_for_cluster_termination(
...
+                        zip(
+                            [res["Instances"][0]["InstanceId"] for res in new_instances],
+                            new_instances,
+                        )
+                    )
+
+                    old_instances = [res for res in previous_instances["Reservations"]]
+                    old_instance_id_to_reservation = dict(
+                        zip(
+                            [res["Instances"][0]["InstanceId"] for res in old_instances],
```
rmoneys, line 894: You could leave this a generate with parens instead of square brackets

tests/test_awsdatabricks.py
```diff
@@ -610,7 +618,106 @@ def test_create_prediction_for_run_success(respx_mock):
...
+                return s3
+            case "ec2":
+                return ec2
+
+    with s3_stubber, ec2_stubber, patch("boto3.client") as mock_aws_client:
+        mock_aws_client.side_effect = client_patch
+        result = create_prediction_for_run("75778", "Premium", "Jobs Compute", "my-project-id")
+
+    assert not result.error
+    assert result.result
```
rmoneys, line 636: The exclusivity of result & error is enforced by a validator in the Response class

rmoneys (approved):
Ignore the musings. Looks good!

sync/awsdatabricks.py
```diff
@@ -218,7 +220,7 @@ def get_cluster_report(
 
 def _get_cluster_report(
     cluster_id: str, plan_type: str, compute_type: str, allow_incomplete: bool
-) -> Response[DatabricksClusterReport]:
+) -> Response[DatabricksClusterReport | dict]:
```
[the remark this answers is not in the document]
gseyffert-sync: why did I add this?.. 🤔

sync/awsdatabricks.py
```diff
@@ -262,6 +248,73 @@ def _get_cluster_report(
...
+                    f"Could not find sync_data/cluster_instances.json for cluster: {cluster_id}"
+                )
+            else:
+                logger.error(
+                    f"Unexpected error encountered while attempting to fetch sync_data/cluster_instances.json: {ex}"
+                )
+
+    # If this cluster does not have the "Sync agent" configured, attempt a best-effort snapshot of the instances that
+    #  are associated with this cluster
+    if cluster_instances is None:
```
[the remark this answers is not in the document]
gseyffert-sync: That's true, though I can't see `cluster_instances` ever being simply `{}`. The reason is because `ec2.describe_instances` will always return a non-empty dict, even if no matching instances are found. The response looks something like -

```python
{
  "Reservations": [],
  "HTTPResponseMetadata": {...}  // This key name be wrong
}
```

And since we are preserving that "shape" during the polling as well, there will always be at least the `Reservations` key in the dict, if it exists.

This is a long way of saying that `if cluster_instances is None` is not really checking for "did we find instances?", but rather for "were we polling during the job execution?"

sync/awsdatabricks.py
```diff
@@ -799,6 +852,89 @@ def _wait_for_cluster_termination(
...
+        (_, bucket, base_prefix) = cluster_log_dest
+
+        aws_region_name = DB_CONFIG.aws_region_name
+        ec2 = boto.client("ec2", region_name=aws_region_name)
+        s3 = boto.client("s3")
+
+        # If the event log destination is just a *bucket* without any sub-path, then we don't want to include
+        #  a leading `/` in our Prefix (which will make it so that we never actually find the event log), so
+        #  we make sure to re-strip our final Prefix
+        file_key = f"{base_prefix}/sync_data/cluster_instances.json".strip("/")
```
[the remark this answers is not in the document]
gseyffert-sync: They aren't required. I think I did this to match the formatting of e.g. the `init_scripts` directory that Databricks writes to S3, but I am definitely open to changing this.

sync/awsdatabricks.py
gseyffert-sync: oh, right you are! Interesting that Python just gives you the keys... what's the point of `.keys()`, then? Indeed, according to the docs, `.keys()` kind of seems superfluous in modern Python -

> iter(d)
      Return an iterator over the keys of the dictionary. This is a shortcut for iter(d.keys()).

https://docs.python.org/3/library/stdtypes.html#typesmapping

tests/test_awsdatabricks.py
gseyffert-sync: ah - yeah, this was a little overzealous asserting, haha

gseyffert-sync:
thanks @rmoneys! I'll include some of your suggestions in #23, if that works :+1:

sync/awsdatabricks.py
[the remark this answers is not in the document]
gseyffert-sync: Ah, I think this is because Pycharm is confused about what is actually being returned in the event of an error -

Screenshot 2023-06-01 at 9 26 36 PM

PR #23, opened 2023-05-30
PROD-820 Expand supported the library's Python versions to [3.7,3.11]

gseyffert-sync:
[PROD-820 Expand supported the library's Python versions to [3.7,3.11]](https://synccomputing.atlassian.net/browse/PROD-820?atlOrigin=eyJpIjoiOTIwMmJmMGFjMTc2NDdlOWIzN2QyMWUyNjdkOTMxNzMiLCJwIjoiaiJ9)

In support of running this library in customer environments, this drops the minimum supported Python version down to 3.7.x

Initially submitting this PR against the `init_scripts` branch in order to get a clean diff, but I can update the target branch to `main` once #22 is approved + merged

The main changes in this PR include -

- Changing `match` statements to `if/elif/else` chains,
- `dict`/`list`/`tuple` generic type-hints to `typing.Dict`/`List`/`Tuple`
- `Union | Types` --> `typing.Union[...]`
- Changing assignment expressions -
```python
if foo := m.get("bar"):
  ...
```
 to -
```python
foo = m.get("bar")
if foo:
  ...
```

### DIFF
--- a/demo/install_cli.sh
+++ b/demo/install_cli.sh
@@ -7,7 +7,7 @@ function find_python() {
   local python_candidate
 
   while read python_candidate; do
-    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3.10(\.|$) ]]; then
+    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3.[7,8,9,10,11](\.|$) ]]; then
       echo $python_candidate
       return
     fi
--- a/demo/install_cli.sh
+++ b/demo/install_cli.sh
@@ -7,7 +7,7 @@ function find_python() {
   local python_candidate
 
   while read python_candidate; do
-    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3.[7,8,9,10,11](\.|$) ]]; then
+    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3\.(7|8|9|10|11)(\.|$) ]]; then
       echo $python_candidate
       return
     fi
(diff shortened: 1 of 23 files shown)


gseyffert-sync:
Tested this in Databricks with DBR 7.3 (the oldest LTS, which comes packaged with Python 3.7.5) and things are looking good there!

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dev = [
   "pytest==7.2.0",
   "pytest-env==0.8.1",
   "pytest-asyncio==0.21.0",
-  "Sphinx==5.3.0",
+  "Sphinx==4.3.0",
```
gseyffert-sync, line 46: Bumped this down because it had some dependencies for `python < 3.10` that were mutually exclusive with `flake8` & `pytest`. Manual doc generation (and poking around those generated docs) seems to work just fine, though

demo/install_cli.sh
```diff
@@ -7,7 +7,7 @@ function find_python() {
   local python_candidate
 
   while read python_candidate; do
-    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3.10(\.|$) ]]; then
+    if [[ $("$python_candidate" --version 2>/dev/null) =~ (^|[^[:digit:].])3.[7,8,9,10,11](\.|$) ]]; then
```
rmoneys, line 10: This isn't quite right, but neither is the original:
```
% [[ "Python 3.17.3" =~ (^|[^[:digit:].])3.[7,8,9,10,11](\.|$) ]] && echo ehhhhh
ehhhhh
```
Instead, I think the pattern should be,
```
(^|[^[:digit:].])3\\.(7|8|9|10|11)(\\.|$)
```

rmoneys (changes requested):

demo/install_cli.sh
rmoneys: nm. 1 back slash is sufficient:
```
(^|[^[:digit:].])3\.(7|8|9|10|11)(\.|$)
```

rmoneys (dismissed):

rmoneys (approved):

PR #24, opened 2023-06-02
Change python version in GH actions

gseyffert-sync:
`3.7.5` is not compatible with Ubuntu 22.04, but `3.7.16` is. Changes from `3.7.5` to `3.7.16` are just security patches, so there is no real functional impact to us.


rmoneys (dismissed):
Probably a good idea to pin the ubuntu version too as in .github/workflows/pull-request.yml so that it doesn't break after a new ubuntu release

rmoneys (approved):

PR #26, opened 2023-06-08
Add DBFS support for downloading event logs

gseyffert-sync:
I think this is "ready" for review for the eventlog portion. Agent/init script logic still needs to be updated to support DBFS, though

### DIFF
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -1,13 +1,14 @@
 """
 Utilities for interacting with Databricks
 """
+import base64
 import io
 import logging
 import time
 import zipfile
 from pathlib import Path
 from time import sleep
-from typing import Any, Dict, List, Tuple, TypeVar, Union, Collection
+from typing import Any, Collection, Dict, List, Tuple, TypeVar, Union
 from urllib.parse import urlparse
 
 import boto3 as boto
@@ -151,6 +152,7 @@ def create_prediction_for_run(
 
     cluster_id_response = _get_run_cluster_id(tasks)
     cluster_id = cluster_id_response.result
+
     if cluster_id:
         # Making these calls prior to fetching the event log allows Databricks a little extra time to finish
         #  uploading all the event log data before we start checking for it
@@ -161,7 +163,8 @@ def create_prediction_for_run(
         if cluster_report:
 
             cluster = cluster_report.cluster
-            eventlog_response = _get_eventlog(cluster, run.get("end_time"))
+            spark_context_id = _get_run_spark_context_id(tasks)
+            eventlog_response = _get_eventlog(cluster, spark_context_id.result, run.get("end_time"))
 
             eventlog = eventlog_response.result
             if eventlog:
@@ -226,7 +229,7 @@ def _get_cluster_report(
 ) -> Response[DatabricksClusterReport]:
     # Cluster `terminated_time` can be a few seconds after the start of the next task in which
     # this may be executing.
-    cluster_response = _wait_for_cluster_termination(cluster_id, timeout_seconds=60, poll_seconds=5)
+    cluster_response = _wait_for_cluster_termination(cluster_id, poll_seconds=5)
     if cluster_response.error:
         return cluster_response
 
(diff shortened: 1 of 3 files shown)


sync/awsdatabricks.py
```diff
@@ -872,22 +860,56 @@ def _wait_for_cluster_termination(
     return Response(error=DatabricksAPIError(**cluster))
 
 
+def _cluster_log_destination(
```
gseyffert-sync, line 863: expanded this function a little bit to handle `dbfs` urls as well.

sync/awsdatabricks.py
```diff
@@ -981,6 +999,18 @@ def _get_run_cluster_id(tasks: List[dict]) -> Response[str]:
...
+def _get_run_spark_context_id(tasks: List[dict]) -> Response[str]:
+    context_ids = {task["cluster_instance"]["spark_context_id"] for task in tasks}
+    num_ids = len(context_ids)
+
+    if num_ids == 1:
+        return Response(result=context_ids.pop())
+    elif num_ids == 0:
+        return Response(error=DatabricksError(message="No cluster found for tasks"))
+    else:
+        return Response(error=DatabricksError(message="More than 1 cluster found for tasks"))
```
gseyffert-sync, line 1011: The `spark_context_id` for a given run is now passed as an argument to `_get_eventlog`. The `spark_context_id` helps us identify, in the event of a shared/re-used cluster, which directory actually contains the eventlog we're interested in. The S3 logic hasn't been updated to use this value yet, but I found it useful for the DBFS directory traversal

sync/awsdatabricks.py
```diff
@@ -1017,71 +1047,205 @@ def _s3_contents_have_all_rollover_logs(contents: List[dict], run_end_time_secon
...
+        False,
+    )
+    return (
+        # Related to - https://docs.databricks.com/clusters/configure.html#cluster-log-delivery-1
+        # We want to make sure that the final_rollover_log has had data written to it AFTER the run has actually
+        # completed. We use this as a signal that Databricks has flushed all event log data to DBFS, and that we
+        # can proceed with the upload + prediction.
+        final_rollover_log
+        and final_rollover_log["modification_time"] >= run_end_time_millis
+    )
```
gseyffert-sync, line 1067: As with S3, check that the most recent event log file has a modification time _after_ our run complete.

sync/awsdatabricks.py
```diff
@@ -1017,71 +1047,205 @@ def _s3_contents_have_all_rollover_logs(contents: List[dict], run_end_time_secon
...
-    log_url = cluster_description.get("cluster_log_conf", {}).get("s3", {}).get("destination")
-    if log_url:
-        parsed_log_url = urlparse(log_url)
+def _get_eventlog_from_s3(
+    cluster_id: str,
+    bucket: str,
+    base_filepath: str,
+    run_end_time_millis: int,
+    poll_duration_seconds: int,
+):
```
gseyffert-sync, line 1081: pulled this out of `_get_eventlog` into its own function now that we also support DBFS

sync/awsdatabricks.py
```diff
@@ -1017,71 +1047,205 @@ def _s3_contents_have_all_rollover_logs(contents: List[dict], run_end_time_secon
...
+        eventlog_file_metadata = eventlog_files.pop()
+        path = eventlog_file_metadata["path"]
+
+        subdir = get_default_client().list_dbfs_directory(path)
+
+        subdir_files = subdir["files"]
+        matching_subdirectory = next(
+            (subfile for subfile in subdir_files if spark_context_id in subfile["path"]),
+            None,
+        )
```
gseyffert-sync, line 1157: Use the `spark_context_id` in order to figure out the correct sub-directory to be looking in

sync/awsdatabricks.py
```diff
@@ -1017,71 +1047,205 @@ def _s3_contents_have_all_rollover_logs(contents: List[dict], run_end_time_secon
...
+            # DBFS tells us exactly how many bytes to expect for each file, so pre-allocate an array
+            #  to write the file chunks in to
+            content = bytearray(filesize)
+            while bytes_read < filesize:
+                file_content = get_default_client().read_dbfs_file_chunk(
+                    eventlog_file_metadata["path"], offset=bytes_read
+                )
+                new_bytes_read = bytes_read + file_content["bytes_read"]
+                content[bytes_read:new_bytes_read] = base64.b64decode(file_content["data"])
+                bytes_read = new_bytes_read
```
gseyffert-sync, line 1196: We can only download files from DBFS in chunks of up to 1 MB - https://docs.databricks.com/api/workspace/dbfs/read. Since Databricks tells us the filesize, this logic just allocates that file size and writes the response chunks into it until we have all the bytes we need.

sync/clients/databricks.py
```diff
@@ -95,6 +95,24 @@ def get_run(self, run_id: str) -> dict:
             self._client.build_request("GET", "/api/2.1/jobs/runs/get", params={"run_id": run_id})
         )
 
+    def list_dbfs_directory(self, path: str) -> dict:
+        return self._send(
+            self._client.build_request("GET", "/api/2.0/dbfs/list", params={"path": path})
+        )
+
+    def read_dbfs_file_chunk(self, path: str, offset: int = 0, length: int = 1024 * 1024) -> dict:
```
gseyffert-sync, line 103: Default `length` to be the maximum value of 1MB

sync/awsdatabricks.py
```diff
@@ -872,22 +860,56 @@ def _wait_for_cluster_termination(
...
+        stripped_path = parsed_log_url.path.strip("/")
+
+        # If the event log destination is just a *bucket* without any sub-path, then we don't want to include
+        #  a leading `/` in our Prefix (which will make it so that we never actually find the event log), so
+        #  we make sure to re-strip our final Prefix
+        base_cluster_filepath_prefix = f"{stripped_path}/{cluster['cluster_id']}".strip("/")
+
+        return log_url, parsed_log_url.scheme, bucket, base_cluster_filepath_prefix
+
+    return None, None, None, None
```
rmoneys, line 884: ParseResult might be a better return type: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.ParseResult
```
>>> p = urlparse("s3://hello/world")
>>> p
ParseResult(scheme='s3', netloc='hello', path='/world', params='', query='', fragment='')
>>> p.geturl()
's3://hello/world'
```

sync/awsdatabricks.py
rmoneys: Interesting that it's not used in S3 eventlog paths

rmoneys (approved):

sync/awsdatabricks.py
gseyffert-sync: It is, if you look at the full path in S3, but we just don't use it (yet)

sync/awsdatabricks.py
gseyffert-sync: I was thinking that after looking at this again... I do want to return `base_cluster_filepath_prefix`, but maybe `Tuple[ParseResult, str]` would be better

PR #28, opened 2023-06-14
Add prediction_params argument to create/update_project functions

gseyffert-sync:
Allow mutating `prediction_params` of a project via the library functions instead of just the UI/API

### DIFF
--- a/sync/api/projects.py
+++ b/sync/api/projects.py
@@ -43,6 +43,7 @@ def create_project(
     description: str = None,
     s3_url: str = None,
     prediction_preference: Preference = Preference.BALANCED,
+    prediction_params: dict = None,
 ) -> Response[dict]:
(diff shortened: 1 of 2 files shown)


sync/api/projects.py
```diff
@@ -54,6 +55,8 @@ def create_project(
     :type s3_url: str, optional
     :param prediction_preference: preferred prediction solution, defaults to Preference.BALANCED
     :type prediction_preference: Preference, optional
+    :param prediction_params: dictionary of prediction parameters, defaults to None. Valid options are documented here - https://developers.synccomputing.com/reference/create_project_v1_projects_post
```
gseyffert-sync, line 58: Seemed better to not try to duplicate API docs in this comment, but I also think this API documentation is getting replaced at some point?.. Or at least, the url may end up changing.

rmoneys (approved):

PR #36, opened 2023-08-09
[PROD-1176] Add Azure support

rmoneys:
https://synccomputing.atlassian.net/browse/PROD-1176

Adds support for Azure Databricks to the library.

Basically, awsdatabricks modules were moved to _databricks and the access report, cluster report and cluster monitoring functions were moved to awsdatabricks and azuredatabricks.

An Azure Databricks cluster report with instances can be had like,
```
sync-cli azure-databricks get-cluster-report 14374
```

For Azure authentication, either set these environment variables:
```
export AZURE_TENANT_ID=069b7263-8ba5-4854-a1dd-4563b01ce1bd
export AZURE_CLIENT_ID=ed551f97-81d3-43a3-8ded-eb4972b67fb1
export AZURE_CLIENT_SECRET=# See the Sync Library app's secrets: https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Credentials/appId/ed551f97-81d3-43a3-8ded-eb4972b67fb1/isMSAApp~/false
export AZURE_SUBSCRIPTION_ID=a3d0a3d4-eabc-4b3e-834b-2dbcdb58ece5
```
Or install and configure the CLI: https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-macos

### DIFF
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.1.5"
+__version__ = "0.1.6"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.1.6"
+__version__ = "0.2.0"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -207,7 +207,13 @@ def _get_cluster_report(
     )
 
 
+if getattr(sync._databricks, "__claim", __name__) != __name__:
+    raise RuntimeError(
+        "Databricks modules for different cloud providers cannot be used in the same context"
+    )
+
 sync._databricks._get_cluster_report = _get_cluster_report
+setattr(sync._databricks, "__claim", __name__)
 
 
 def _get_cluster_instances(cluster: dict) -> Response[dict]:
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -365,8 +365,8 @@ def _monitor_cluster(
             else:
                 # Otherwise, update old references and include any new instances in the list
                 newly_added_instance_ids = new_instance_ids.difference(old_instance_ids)
-                updated_instance_ids = newly_added_instance_ids.intersection(old_instance_ids)
-                removed_instance_ids = old_instance_ids.difference(updated_instance_ids)
+                updated_instance_ids = new_instance_ids.intersection(old_instance_ids)
+                removed_instance_ids = old_instance_ids.difference(new_instance_ids)
 
                 removed_instances = [
                     old_instance_id_to_reservation[id] for id in removed_instance_ids
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -243,7 +243,7 @@ def _get_cluster_instances(cluster: dict) -> Response[dict]:
 
     # If this cluster does not have the "Sync agent" configured, attempt a best-effort snapshot of the instances that
     #  are associated with this cluster
-    if cluster_instances is None:
+    if not cluster_instances:
         ec2 = boto.client("ec2", region_name=aws_region_name)
         try:
             cluster_instances = ec2.describe_instances(
--- a/sync/azuredatabricks.py
+++ b/sync/azuredatabricks.py
@@ -1,9 +1,11 @@
 import logging
 import os
 from time import sleep
+from typing import Type, TypeVar
 from urllib.parse import urlparse
 
 import orjson
+from azure.common.credentials import get_cli_profile
 from azure.core.exceptions import ClientAuthenticationError
 from azure.identity import DefaultAzureCredential
 from azure.mgmt.compute import ComputeManagementClient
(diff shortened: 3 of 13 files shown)


sync/awsdatabricks.py
```diff
@@ -353,28 +207,7 @@ def _get_cluster_report(
...
-
-
-def _get_cluster_instances_from_dbfs(filepath: str):
-    filepath = format_dbfs_filepath(filepath)
-    dbx_client = get_default_client()
-    try:
-        return read_dbfs_file(filepath, dbx_client)
-    except Exception as ex:
-        logger.error(f"Unexpected error encountered while attempting to fetch {filepath}: {ex}")
+sync._databricks._get_cluster_report = _get_cluster_report
```
gseyffert-sync, line 210: Interesting. How does this work if you try to use both the Azure and AWS modules? This kind of seems like you get one or the other

sync/__init__.py
```diff
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.1.5"
+__version__ = "0.1.6"
```
gseyffert-sync, line 2: `0.2.0`, maybe?

sync/azuredatabricks.py
```diff
@@ -0,0 +1,411 @@
...
+    if not cluster_instances:
+
+        resource_group_name = _get_databricks_resource_group_name()
+
+        compute = _get_azure_client(ComputeManagementClient)
+        if resource_group_name:
+            vms = compute.virtual_machines.list(resource_group_name=resource_group_name)
+        else:
+            logger.warning("Failed to find Databricks managed resource group")
+            vms = compute.virtual_machines.list_all()
```
gseyffert-sync, line 260: are these responses not paginated? The docs are a little hard to grok, but I see a `next` field in the example response here - https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/list?tabs=HTTP

Having to go through all VMs in a workspace seems like it will be slow... is there really no way to query based on tags?

sync/awsdatabricks.py
```diff
@@ -1155,8 +365,8 @@ def write_file(body: bytes):
             else:
                 # Otherwise, update old references and include any new instances in the list
                 newly_added_instance_ids = new_instance_ids.difference(old_instance_ids)
-                updated_instance_ids = newly_added_instance_ids.intersection(old_instance_ids)
-                removed_instance_ids = old_instance_ids.difference(updated_instance_ids)
+                updated_instance_ids = new_instance_ids.intersection(old_instance_ids)
+                removed_instance_ids = old_instance_ids.difference(new_instance_ids)
```
rmoneys, line 369: Looks like there was a bug here

sync/azuredatabricks.py
```diff
@@ -0,0 +1,411 @@
...
+
+    resource_group_name = _get_databricks_resource_group_name()
+    if not resource_group_name:
+        logger.warning("Failed to find Databricks managed resource group")
+
+    previous_instances = {}
+
+    while True:
+        try:
+            compute = _get_azure_client(ComputeManagementClient)
```
gseyffert-sync, line 325: nit - any reason for this to be inside the `while`?

sync/azuredatabricks.py
```diff
@@ -0,0 +1,411 @@
...
+            logger.error(f"Exception encountered while polling cluster: {e}")
+
+        sleep(polling_period)
+
+
+def _get_databricks_resource_group_name() -> str:
+    resources = _get_azure_client(ResourceManagementClient)
+
+    for workspace_item in resources.resources.list(
+        filter="resourceType eq 'Microsoft.Databricks/workspaces'"
```
gseyffert-sync, line 387: love the weird query language, haha. Thanks Microsoft

sync/awsdatabricks.py
gseyffert-sync: Oops - yep, this changes looks correct to me.

sync/awsdatabricks.py
```diff
@@ -1,36 +1,84 @@
...
-    DatabricksClusterReport,
+    AWSDatabricksClusterReport,
     DatabricksError,
-    Platform,
     Response,
 )
-from sync.utils.dbfs import format_dbfs_filepath, read_dbfs_file, write_dbfs_file
+from sync.utils.dbfs import format_dbfs_filepath, write_dbfs_file
+
+__all__ = [
```
rmoneys, line 53: This is required for the functions to show up in the documentation

gseyffert-sync (commented):
Mostly LGTM, but I have a few questions. In general, pretty great that most of the Databricks logic could be re-used :+1:

sync/awsdatabricks.py
rmoneys: Yeah, good call-out. You can't use both modules together. I don't love the patching here, but I couldn't see a better way without duplicating a bunch of code - I should've encapsulated the API in classes rather than modules at the start. For now though, perhaps a worthwhile guard is to raise an error if someone tries to import both.

Configuration only allows 1 Databricks workspace so there's no reason to import both.

sync/__init__.py
rmoneys: I can dig it

sync/azuredatabricks.py
rmoneys: Yeah, I'm kinda concerned about that too. After some searching I concluded that you really can't filter by tags 🤦

I think the paging goes on under the hood as you iterate through the result - I didn't come across any paging parameters, but I noted that the request doesn't happen until you start pulling vms out of `vms`.

sync/azuredatabricks.py
rmoneys: Nope - I had meant to move that up.

sync/azuredatabricks.py
rmoneys: Yeah, and that it's partially implemented on just about every API is really great too. They love to keep you guessing.

sync/awsdatabricks.py
gseyffert-sync: Yeah, there are definitely some practical limitations to using both modules, so I don't think this is a "blocking" issue, but yeah, a basic guardrail might be good just in case

sync/azuredatabricks.py
gseyffert-sync: > After some searching I concluded that you really can't filter by tags 🤦

Oof. Something to watch out for, for sure

> I think the paging goes on under the hood as you iterate through the result

Ah, very possible! Yeah, I was looking at the REST reference, not the Python SDK (docs on the Python SDK seem quite hard to find??)

gseyffert-sync (approved):

PR #37, opened 2023-08-15
add ebs volume info to the cluster report

gorskysd:
Add ebs volume information to the cluster report. Information is gathered via the `describe_volumes` call. This is necessary because the `describe_instances` call does not provide comprehensive information about any attached ebs storage, including the storage type, IOPS, and throughput.

To enable reusing most of the existing code without refactor I added the list of volumes as an attribute in the `instances` dict. I considered break this dict apart in the cluster report for a sense of "separation of duties", but I wasn't sure if that was any better. For example:
```
class AWSDatabricksClusterReport(DatabricksClusterReport):
    instances: Union[dict, None]
    volumes: Union[dict, None] #  <--- new field for volumes?
```

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -223,7 +223,7 @@ def _get_cluster_report(
     raise NotImplementedError()
 
 
-def _get_cluster_instances_from_dbfs(filepath: str):
+def _get_aws_cluster_info_from_dbfs(filepath: str):
     filepath = format_dbfs_filepath(filepath)
     dbx_client = get_default_client()
     try:
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -223,7 +223,7 @@ def _get_cluster_report(
     raise NotImplementedError()
 
 
-def _get_aws_cluster_info_from_dbfs(filepath: str):
+def _get_cluster_instances_from_dbfs(filepath: str):
     filepath = format_dbfs_filepath(filepath)
     dbx_client = get_default_client()
     try:
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -29,6 +29,7 @@ def create_prediction(
     cluster: dict,
     cluster_events: dict,
     instances: dict,
+    volumes: dict,
     eventlog: bytes,
     project_id: str = None,
 ) -> Response[str]:
@@ -68,6 +69,7 @@ def create_prediction(
             "cluster": cluster,
             "cluster_events": cluster_events,
             "instances": instances,
+            "volumes": volumes,
         },
         "eventlog.zip",
         eventlog,
@@ -167,6 +169,7 @@ def create_prediction_for_run(
                     cluster=cluster,
                     cluster_events=cluster_report.cluster_events,
                     instances=cluster_report.instances,
+                    volumes=cluster_report.volumes,
                     eventlog=eventlog,
                     project_id=project_id,
                 )
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -12,7 +12,7 @@ import sync._databricks
 from sync._databricks import (
     _cluster_log_destination,
     _get_all_cluster_events,
-    _get_cluster_instances_from_dbfs,
+    _get_aws_cluster_info_from_dbfs,
     _wait_for_cluster_termination,
     create_and_record_run,
     create_and_wait_for_run,
@@ -190,9 +190,9 @@ def _get_cluster_report(
 
     cluster = cluster_response.result
 
-    instances, volumes = _get_cluster_instances(cluster)
+    reservations, volumes = _get_aws_cluster_info(cluster)
 
-    for response in [instances, volumes]:
+    for response in [reservations, volumes]:
         if response.error:
             if allow_incomplete:
                 logger.warning(response.error)
@@ -206,7 +206,7 @@ def _get_cluster_report(
             compute_type=compute_type,
             cluster=cluster,
             cluster_events=cluster_events,
-            instances=instances.result,
+            instances=reservations.result,
             volumes=volumes.result,
         )
     )
@@ -221,8 +221,8 @@ sync._databricks._get_cluster_report = _get_cluster_report
 setattr(sync._databricks, "__claim", __name__)
 
 
-def _get_cluster_instances(cluster: dict) -> Tuple[Response[dict], Response[dict]]:
-    cluster_instances = None
+def _get_aws_cluster_info(cluster: dict) -> Tuple[Response[dict], Response[dict]]:
+    cluster_info = None
     aws_region_name = DB_CONFIG.aws_region_name
 
     cluster_log_dest = _cluster_log_destination(cluster)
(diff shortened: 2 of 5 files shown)


sync/awsdatabricks.py
```diff
@@ -308,7 +307,7 @@ def monitor_cluster(cluster_id: str, polling_period: int = 30) -> None:
         logger.warning("Unable to monitor cluster due to missing cluster log destination - exiting")
 
 
-def _monitor_cluster(
+def _monitor_cluster(  # noqa: C901
```
gorskysd, line 310: Obviously not ideal, but this addition barely increased the complexity over the line, and I didn't see any obvious factoring that would *actually* make this logic clearer to understand.

sync/awsdatabricks.py
```diff
@@ -388,10 +381,69 @@ def write_file(body: bytes):
...
+
+        if new_instances := response.get("Reservations"):
+            instances += new_instances
+
+        if next_token := response.get("NextToken"):
+            args["NextToken"] = next_token
+        else:
+            break
+
+    return {"Reservations": instances}
```
gorskysd, line 449: The `describe_volumes` call is new. I also factored out the `describe_instances` call because it was common to two places, and it was a good opportunity to beef it up with some NextToken handling for very large clusters.

sync/awsdatabricks.py
```diff
@@ -388,10 +382,67 @@ def write_file(body: bytes):
...
-            write_file(orjson.dumps(instances))
+            # Now record the ebs volumes
+            volumes = _get_ebs_volumes(cluster_id, ec2)
+
+            for v in volumes:
+                if v["VolumeId"] not in recorded_volume_id_set:
+                    recorded_volume_id_set.add(v["VolumeId"])
+                    recorded_volumes.append(v)
 
+            instances["Volumes"] = recorded_volumes
```
rmoneys, line 393: This solution won't extend to Azure Databricks because the instances in a cluster report is an array instead of a dict: https://github.com/synccomputingcode/syncsparkpy/blob/c5575c873acc150e9f5d931f0adca0c8296d2122/sync/models.py#L110-L111
We could make it a dict, but seems better to put volumes under "volumes" at the top level if only for better naming.

gorskysd: Gotcha, that's something I was considering doing, and I think I like it more. It maintains some parity with Azure (though idk if a separate call like this is needed for Azure), but it also keeps the existing structures unchanged which has some benefit to mental load.

rmoneys: You said it better than me!

sync/awsdatabricks.py
```diff
@@ -388,10 +382,67 @@ def write_file(body: bytes):
...
                 }
 
-            write_file(orjson.dumps(instances))
+            # Now record the ebs volumes
+            volumes = _get_ebs_volumes(cluster_id, ec2)
+
+            for v in volumes:
+                if v["VolumeId"] not in recorded_volume_id_set:
+                    recorded_volume_id_set.add(v["VolumeId"])
+                    recorded_volumes.append(v)
```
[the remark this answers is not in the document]
rmoneys: Hopefully that'll remove the need for the noqa directive

sync/awsdatabricks.py
```diff
@@ -388,10 +382,67 @@ def write_file(body: bytes):
...
             logger.error(f"Exception encountered while polling cluster: {e}")
 
         sleep(polling_period)
+
+
+def _get_ebs_volumes(cluster_id: str, ec2_client: "botocore.client.ec2") -> List[Dict]:
+
+    args = {}
+    volumes = []
+    while True:
```
gseyffert-sync, line 407: For both this function and `_get_ec2_instances`, I personally prefer a pattern more like -

```python
response = ec2_client.describe_volumes(...)
volumes = response.get("Volumes", [])
next_token = response.get("NextToken")

while next_token:
    response = ec2_client.describe_volumes(..., NextToken=next_token)
    volumes += response.get("Volumes", [])
    next_token = response.get("NextToken")

return volumes
```

This eliminates the need for a `break` in your loop, and explicitly separates the initial load logic from the paging logic

gseyffert-sync: I will add that I used to always try to put the initial / paging logic in the same loop, but I've given up on that. Separating out the two now generally, I find, leads to code that is easier to follow / update

sync/awsdatabricks.py
```diff
@@ -1,8 +1,10 @@
 import logging
 from time import sleep
+from typing import Dict, List
```
gseyffert-sync, line 3: Noticed you initially were using `dict`/`list` - just wanted to explicitly state, in case you were like "wft?", that this is for Python 3.7 support to be compatible with older Databricks versions :+1:

sync/awsdatabricks.py
gorskysd: Happy to make this change. Just curious, does loop breaking like this make you nervous from a brittleness standpoint?

sync/awsdatabricks.py
gorskysd: Yeeeep. I remember when we upgraded from 3.7 to 3.10. I knew what was up right away.

sync/awsdatabricks.py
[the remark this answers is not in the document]
gorskysd: Good call, noqa is gone

gorskysd:
Thanks for the feedback, y'all. In the process of making "volumes" a higher tier entity I broke some tests. Fixed those, and added a mock for the `describe_volumes` call.

I tested the changes on both an EBS cluster and a storage attached NVMe cluster (though the latter still has two ebs volumes attached per instance by default).

sync/awsdatabricks.py
```diff
@@ -216,7 +221,7 @@ def _get_cluster_report(
 setattr(sync._databricks, "__claim", __name__)
 
 
-def _get_cluster_instances(cluster: dict) -> Response[dict]:
+def _get_cluster_instances(cluster: dict) -> Tuple[Response[dict], Response[dict]]:
```
gseyffert-sync, line 224: nit - Perhaps worth re-naming this function to something like `_get_cluster_aws_information`?

gseyffert-sync (approved):
I think this LGTM. Just to make sure I read it correctly, the top-level file structure that we write in `monitor_cluster` will be something like -

```javascript
{
  "Reservations": [...],
  "Volumes": [...],
}
```

Where `"Reservations"`/`"Volumes"`` are both top-level keys from the corresponding AWS `describe_*` calls?

Since we are doing this for more than just instances now, we can probably remove some of the intermediate `"Reservations"` usages that are sprinkled around here, but I can also handle that in some upcoming work I need to do

gorskysd:
@gseyffert-sync on second thought, those nits are rather important for clarity down the line. 

I did a whole bunch of relabeling to avoid conflating response data from `describe_instances` ("Reservations") and `describe_volumes` ("Volumes"). These two objects do get combined in the file and in the cluster report as:
```
# In saved file
aws_cluster_info = {
    "Reservations" : ...
    "Volumes" : ...
}

# In cluster report
AWSDatabricksClusterReport(
    instances : {"Reservations" : [...]}
    volumes: {"Volumes" : [...]}
)
```

A few other things that changed:

- (!!) Changed the stored filename to `aws_cluster_info.json`. This was for consistency points. If you think it will be problematic I'll change it back!
- A few methods had their definitions changed to use `aws_cluster_info`. EXCEPT for `_get_cluster_instances_from_dbfs` imported from `_databricks.py`.
- Made `_get_ec2_instances` and `_get_ebs_volumes` symmetric such that they both remove the top-tier field and just return the list instead. This allowed for a lot of simplification in `_monitor_cluster`. 
    * The top-tier fields are reintroduced when saving the file, and both are persisted in the cluster report as well. 
    * Obviously the double nestings in the cluster report are unnecessary, but I'd rather treat these two AWS entities symmetrically throughout the library, and also not change what the api expects for `instances`.

sync/_databricks.py
```diff
@@ -223,7 +223,7 @@ def _get_cluster_report(
     raise NotImplementedError()
 
 
-def _get_cluster_instances_from_dbfs(filepath: str):
+def _get_aws_cluster_info_from_dbfs(filepath: str):
     filepath = format_dbfs_filepath(filepath)
     dbx_client = get_default_client()
     try:
```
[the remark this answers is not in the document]
gseyffert-sync: sorry, I tried to cancel my review - these comments are no longer relevant

gseyffert-sync (dismissed):

gseyffert-sync:
Looks good to me! We will need to add this `volumes` field to the API model as well, I'd imagine

sync/awsdatabricks.py
```diff
@@ -227,64 +232,67 @@ def _get_cluster_instances(cluster: dict) -> Response[dict]:
...
         if ex.response["Error"]["Code"] == "NoSuchKey":
             logger.warning(
-                f"Could not find sync_data/cluster_instances.json for cluster: {cluster_id}"
+                f"Could not find sync_data/aws_cluster_info.json for cluster: {cluster_id}"
             )
         else:
             logger.error(
-                f"Unexpected error encountered while attempting to fetch sync_data/cluster_instances.json: {ex}"
+                f"Unexpected error encountered while attempting to fetch sync_data/aws_cluster_info.json: {ex}"
             )
```
rmoneys, line 296: Potential improvement:
```
except ClientError as err:
    logger.warning(f"Failed to retrieve cluster info from S3 with key, '{file_key}': {err}")
```
The reason for the exception is not important to the consumer which anticipates a `None` response so it should be a warning in either case. If we need to better understand the reason it'll be in the log statement, e.g. "Failed to retrieve cluster info from S3 with key, 'sync_data/1234210239561239/aws_cluster_info.json': An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist."

rmoneys (commented):
Looks good, but our web API won't see volume info until we make some changes further up the call hierarchy. sync._databricks.create_prediction_for_run ignores the "volumes" attribute on the AWSDatabricksClusterReport - do we want use volume info in the current prediction model?

sync/awsdatabricks.py
gorskysd: Done! Happy to make changes that cut out lines :D

gorskysd:
Originally wasn't gonna connect this to the API since it isn't really needed with the current predictor version, but I might as well just do it and make the backend changes now. I'll spin up a sister branch for the backend.

rmoneys (dismissed):
You may also want to update the version in sync.__info__.py if you want to make this available to users, but approving in case you think this is a good unit of work as is.

gseyffert-sync (approved):

PR #38, opened 2023-08-23
Graham/create submissions

gseyffert-sync:
This PR adds support for submitting data to the `/projects/{id}/submissions` endpoint for Databricks (AWS & Azure) as well as. Also adds support for passing the `tasks` that are related to a given Sync Project in the `/submissions` payload (there will be a corresponding backend PR for that).

After this PR is merged I'll quickly submit a 2nd one to add support for `/recommendations`

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -17,13 +17,7 @@ from sync.api.predictions import create_prediction_with_eventlog_bytes, get_pred
 from sync.api.projects import create_project_submission_with_eventlog_bytes, get_project
 from sync.clients.databricks import get_default_client
 from sync.config import CONFIG
-from sync.models import (
-    AWSDatabricksClusterReport,
-    DatabricksAPIError,
-    DatabricksClusterReport,
-    DatabricksError,
-    Response,
-)
+from sync.models import DatabricksAPIError, DatabricksClusterReport, DatabricksError, Response
 from sync.utils.dbfs import format_dbfs_filepath, read_dbfs_file
 
 logger = logging.getLogger(__name__)
@@ -37,6 +31,7 @@ def create_prediction(
     instances: dict,
     eventlog: bytes,
     volumes: dict = None,
+    tasks: List[dict] = None,
     project_id: str = None,
 ) -> Response[str]:
     """Create a Databricks prediction
@@ -62,6 +57,8 @@ def create_prediction(
         function
     :param eventlog: encoded event log zip
     :type eventlog: bytes
+    :param tasks: The Databricks Tasks associated with the cluster
+    :type tasks: List[dict]
     :param volumes: The EBS volumes that were attached to this cluster
     :type volumes: dict, optional
     :param project_id: Sync project ID, defaults to None
@@ -78,6 +75,7 @@ def create_prediction(
             "cluster_events": cluster_events,
             "instances": instances,
             "volumes": volumes,
+            "tasks": tasks,
         },
         "eventlog.zip",
         eventlog,
@@ -156,7 +154,7 @@ def create_prediction_for_run(
 
     cluster_report, eventlog = run_information_response.result
     return create_prediction(
-        **cluster_report.dict(),
+        **cluster_report.dict(exclude_none=True),
         eventlog=eventlog,
         project_id=project_id,
     )
@@ -170,6 +168,7 @@ def create_submission(
     cluster_events: dict,
     instances: dict,
     eventlog: bytes,
+    tasks: List[dict] = None,
     volumes: dict = None,
 ) -> Response[str]:
     """Create a Databricks Project submission
@@ -197,9 +196,11 @@ def create_submission(
         function
     :param eventlog: encoded event log zip
     :type eventlog: bytes
+    :param tasks: The Databricks Tasks associated with the cluster
+    :type tasks: List[dict]
     :param volumes: The EBS volumes that were attached to this cluster
     :type volumes: dict, optional
-    :return: prediction ID
+    :return: Submission ID
     :rtype: Response[str]
     """
     return create_project_submission_with_eventlog_bytes(
@@ -211,6 +212,7 @@ def create_submission(
             "cluster_events": cluster_events,
             "instances": instances,
             "volumes": volumes,
+            "tasks": tasks,
         },
         "eventlog.zip",
         eventlog,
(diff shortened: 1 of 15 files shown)


sync/_databricks.py
```diff
@@ -887,10 +1031,33 @@ def _get_job_cluster(tasks: List[dict], job_clusters: list) -> Response[dict]:
...
+):
+    tasks = [
+        task for task in all_tasks if not exclude_tasks or task["task_key"] not in exclude_tasks
+    ]
+    if project_id:
+        project_tasks = [
+            t
+            for t in tasks
+            if t.get("new_cluster", {}).get("custom_tags", {}).get("sync:project-id") == project_id
+        ]
```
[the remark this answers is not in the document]
rmoneys: Looks like this could potentially get us tasks that run on different clusters. And what about the tasks that run on a job cluster?

What if we create a mapping of clusters to tasks here with only the non-excluded tasks. Filter the dict by project ID if we have one. Error if we end up with an empty dict or one with more than 1 item.

Maybe we move this all into `_get_run_cluster_id` and rename it - it'll return that 1 dict item or error. for the job clusters we'll have to pull it down in a separate API transaction so we may want a `_get_cluster_report` that takes a cluster object and have the current `_get_cluster_report` call it after getting a cluster if we still need that interface.

sync/clients/sync.py
```diff
@@ -13,7 +13,8 @@ class SyncAuth(httpx.Auth):
     requires_response_body = True
 
     def __init__(self, api_url: str, api_key: APIKey):
-        self.auth_url = f"{api_url}/v1/auth/token"
+        # self.auth_url = f"{api_url}/v1/auth/token"
+        self.auth_url = "https://dev-api.synccomputing.com/v1/auth/token"
```
rmoneys, line 17: This can be achieved by setting `api_url` in ~/.sync/config

sync/_databricks.py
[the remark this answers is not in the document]
gseyffert-sync: > Looks like this could potentially get us tasks that run on different clusters.

This is true, though in practice this is enforced when we call `_get_run_cluster_id` in order to get the cluster info.

> Maybe we move this all into _get_run_cluster_id and rename it - it'll return that 1 dict item or error.

Yeah, I think this is a good idea :+1:

> And what about the tasks that run on a job cluster?

Do you mean when we're applying a recommendation to a Job Definition? As far as I'm aware, we don't distinguish between e.g. All-Purpose and Jobs Compute clusters when creating a submission 🤔

sync/clients/sync.py
gseyffert-sync: Ah, yeah, I did not mean to commit this 😅  I had my config pointed to local but wanted the auth call to point to dev, since fetching a token from my local address was causing me issues for some reason

tests/api/test_predictions.py
```diff
@@ -3,10 +3,11 @@
 from httpx import Response
 
 from sync.api import predictions
+from sync.config import CONFIG
 from sync.models import Platform
 
 # auth route will only be called in the first test
-mock_router = respx.mock(base_url="https://*.synccomputing.com", assert_all_called=False)
+mock_router = respx.mock(base_url=CONFIG.api_url, assert_all_called=False)
```
gseyffert-sync, line 10: This started failing when running the tests pointed to my local API instance. Figured there was no harm in using this config value for the tests

sync/_databricks.py
```diff
@@ -885,16 +1026,41 @@ def _get_job_cluster(tasks: List[dict], job_clusters: list) -> Response[dict]:
...
+    all_tasks: List[dict],
+    exclude_tasks: Union[Collection[str], None] = None,
+    project_id: str = None,
+) -> Tuple[str, List[dict]]:
+    cluster_id_to_tasks = {}
+    for task in all_tasks:
+        if not exclude_tasks or task["task_key"] not in exclude_tasks:
+            cluster_id = task["cluster_instance"]["cluster_id"]
+            if cluster_id not in cluster_id_to_tasks:
+                cluster_id_to_tasks[cluster_id] = [task]
```
rmoneys, line 1039: Looks like a good place to use `defaultdict(list)`

sync/_databricks.py
```diff
@@ -885,16 +1026,41 @@ def _get_job_cluster(tasks: List[dict], job_clusters: list) -> Response[dict]:
...
+            else:
+                cluster_id_to_tasks[cluster_id].append(task)
+
+    num_ids = len(cluster_id_to_tasks)
+    if num_ids == 0:
+        raise Exception("No cluster found for tasks")
+    elif num_ids > 1:
+        raise Exception("More than 1 cluster found for tasks")
```
rmoneys, line 1047: Can we move this to after we optionally filter by project ID? That way we can target clusters within a job.

sync/_databricks.py
```diff
@@ -885,16 +1026,41 @@ def _get_job_cluster(tasks: List[dict], job_clusters: list) -> Response[dict]:
...
+
+    cluster_id, tasks = cluster_id_to_tasks.popitem()
+    if project_id:
+        project_tasks = [
+            t
+            for t in tasks
+            if t.get("new_cluster", {}).get("custom_tags", {}).get("sync:project-id") == project_id
```
rmoneys, line 1054: Tasks that use a cluster defined in the job won't have a "new_cluster" item. I think we'll need to pull down the cluster for each cluster ID. For example, this run, https://dbc-d95d06ca-1d00.cloud.databricks.com/?o=656201176161048#job/769372317240466/run/14244803, that only used the job cluster:
```
>>> print(orjson.dumps(databricks.get_default_client().get_run(14244803)['tasks'], option=orjson.OPT_INDENT_2).decode())
[
  {
    "run_id": 14245362,
    "task_key": "load_all_lending_tables",
    "run_if": "ALL_SUCCESS",
    "notebook_task": {
      "notebook_path": "/Repos/scott.romney@synccomputing.com/lendingclub/lending_club_medallion_full",
      "source": "WORKSPACE"
    },
    "job_cluster_key": "Job_cluster",
    "state": {
      "life_cycle_state": "TERMINATED",
      "result_state": "SUCCESS",
      "state_message": "",
      "user_cancelled_or_timedout": false
    },
    "run_page_url": "https://dbc-d95d06ca-1d00.cloud.databricks.com/?o=656201176161048#job/769372317240466/run/14245362",
    "start_time": 1693617683584,
    "setup_duration": 312000,
    "execution_duration": 235000,
    "cleanup_duration": 0,
    "end_time": 1693618230789,
    "cluster_instance": {
      "cluster_id": "0902-012124-csylezvn",
      "spark_context_id": "8442236568853351267"
    },
    "attempt_number": 0
  }
]
```

rmoneys (changes requested):
I don't think this PR correctly filters clusters/tasks by project ID, and I think we don't want to fail when getting a cluster report if we have more than 1 cluster in a run, but only 1 is tagged with the provided project ID.

sync/cli/awsemr.py
```diff
@@ -102,6 +102,23 @@ def create_project_prediction(project: Dict[str, str], run_id: str = None, regio
         click.echo(f"Failed to create prediction. {prediction_response.error}", err=True)
 
 
+@aws_emr.command
+@click.option(
+    "--project", callback=validate_project, help="The project ID for which to submit data."
+)
```
rmoneys, line 108: This is required isn't it?

gseyffert-sync: Ah, good catch. I made it an `argument` for Databricks

rmoneys (approved):

PR #41, opened 2023-09-15
Sean/better ebs handling

gorskysd:
Changed EBS fetching to filter by attached instance-ids rather than the `ClusterId` tag. Turns out databricks will repurpose EBS volumes, but WILL NOT update the tags when it does so. Harumph.

Also fixed an iteration bug on the tasks that would occasionally throw an error.

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -277,6 +277,7 @@ def _get_run_information(
     exclude_tasks: Union[Collection[str], None] = None,
 ) -> Response[Tuple[DatabricksClusterReport, bytes]]:
     run = get_default_client().get_run(run_id)
+
     if "error_code" in run:
         return Response(error=DatabricksAPIError(**run))
 
@@ -1031,6 +1032,7 @@ def _get_cluster_id_and_tasks_from_run_tasks(
     job_clusters = {c["job_cluster_key"]: c["new_cluster"] for c in run.get("job_clusters", [])}
     project_cluster_ids = defaultdict(list)
     all_cluster_tasks = defaultdict(list)
+
     for task in run["tasks"]:
         if not exclude_tasks or task["task_key"] not in exclude_tasks:
             cluster_id = task["cluster_instance"]["cluster_id"]
@@ -1051,7 +1053,7 @@ def _get_cluster_id_and_tasks_from_run_tasks(
         if cluster_ids:
             project_cluster_tasks = {
                 cluster_id: tasks
-                for cluster_id, tasks in all_cluster_tasks
+                for cluster_id, tasks in all_cluster_tasks.items()
                 if cluster_id in cluster_ids
             }
         else:
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -283,6 +283,7 @@ def _get_aws_cluster_info(cluster: dict) -> Tuple[Response[dict], Response[dict]
     if not cluster_info:
         volumes_response = Response(error=DatabricksError(message=missing_message("ebs volumes")))
     else:
+        logger.warning(missing_message("ebs volumes"))
         volumes_response = Response(result={"Volumes": cluster_info.get("Volumes", [])})
 
     return reservations_response, volumes_response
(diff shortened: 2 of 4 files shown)


sync/awsdatabricks.py
```diff
@@ -280,10 +284,11 @@ def missing_message(input: str) -> str:
     else:
         reservations_response = Response(result={"Reservations": cluster_info["Reservations"]})
 
-    if not cluster_info or not cluster_info["Volumes"]:
+    if not cluster_info:
         volumes_response = Response(error=DatabricksError(message=missing_message("ebs volumes")))
     else:
-        volumes_response = Response(result={"Volumes": cluster_info["Volumes"]})
+        logger.warning(missing_message("ebs volumes"))
+        volumes_response = Response(result={"Volumes": cluster_info.get("Volumes", [])})
```
gorskysd, line 291: It seems silly to return an error if there isn't ebs data. Instead just give a warning message.

sync/_databricks.py
```diff
@@ -1051,7 +1053,7 @@ def _get_cluster_id_and_tasks_from_run_tasks(
         if cluster_ids:
             project_cluster_tasks = {
                 cluster_id: tasks
-                for cluster_id, tasks in all_cluster_tasks
+                for cluster_id, tasks in all_cluster_tasks.items()
```
gorskysd, line 1056: iterating over a dict only iterates through the keys, so this would throw an error.  The `.items()` method must be called to return key-value pairs as a tuple.

sync/awsdatabricks.py
```diff
@@ -435,4 +424,43 @@ def _get_ec2_instances(cluster_id: str, ec2_client: "botocore.client.ec2") -> Li
...
+        for res in reservations:
+            for instance in res.get("Instances", []):
+                if instance:
+                    instance_ids.append(instance.get("InstanceId"))
+
+    volumes = []
+    if instance_ids:
+        filters = [
+            {"Name": "tag:Vendor", "Values": ["Databricks"]},
+            {"Name": "attachment.instance-id", "Values": instance_ids},
```
gseyffert-sync, line 451: :+1:

gseyffert-sync (approved):

gseyffert-sync (commented):
Probably worth bumping `__version__` as well

sync/awsdatabricks.py
```diff
@@ -273,17 +277,17 @@ def missing_message(input: str) -> str:
...
         )
     else:
         reservations_response = Response(result={"Reservations": cluster_info["Reservations"]})
 
-    if not cluster_info or not cluster_info["Volumes"]:
+    if not cluster_info:
         volumes_response = Response(error=DatabricksError(message=missing_message("ebs volumes")))
     else:
-        volumes_response = Response(result={"Volumes": cluster_info["Volumes"]})
+        volumes_response = Response(result={"Volumes": cluster_info.get("Volumes", [])})
```
rmoneys, line 290: Just above we're returning an error for missing volume information. Volume information is potentially missing here as well - shouldn't these 2 clauses be consistent?

gorskysd: For the current product getting the EBS information isn't critical, so I decided to not have it throw an error if ONLY the ebs volume data is missing. If the cluster info is missing entirely then an error object still needs to be returned for both EBS & Reservations, which is why the logic is the way it is nowl

rmoneys: Ok. For instances return an error whenever we have no instance info to return. For volumes return an error only when both volume and instance info are missing? The volume part of the tuple response seems to be representing more than just the state of volume info. I think it's easier to wrap your head around the behavior of this function if each response object is only concerned with the data that it'd contain in a successful response. I understand volume info isn't as critical as instance info - would returning an error for missing volume info break anything? This method is not part of the developer interface however, so this isn't going to hold up the PR.

rmoneys (approved):

sync/awsdatabricks.py
gorskysd: Good point, I can massage this a little bit more to make it more clear.

gorskysd: Yeah that's a good call. I moved the error handling up to the `_get_cluster_report`, where it makes a lot more sense to be.

rmoneys (approved):

PR #42, opened 2023-09-19
ClI update n' bug fix

rmoneys:


sync/_databricks.py
```diff
@@ -29,8 +29,8 @@ def create_prediction(
     compute_type: str,
     cluster: dict,
     cluster_events: dict,
-    instances: dict,
     eventlog: bytes,
+    instances: dict = None,
```
rmoneys, line 33: We still need to be able to run with incomplete reports

sync/_databricks.py
```diff
@@ -1051,7 +1052,7 @@ def _get_cluster_id_and_tasks_from_run_tasks(
         if cluster_ids:
             project_cluster_tasks = {
                 cluster_id: tasks
-                for cluster_id, tasks in all_cluster_tasks
+                for cluster_id, tasks in all_cluster_tasks.items()
```
rmoneys, line 1055: This was causing an error:
```
too many values to unpack (expected 2)
```

sync/api/projects.py
```diff
@@ -102,6 +102,7 @@ def update_project(
     project_id: str,
     description: str = None,
     s3_url: str = None,
+    app_id: str = None,
```
rmoneys, line 105: I find it useful to be able to set the external ID so that the same ID can be used across environments

sync/_databricks.py
```diff
@@ -1032,7 +1033,9 @@ def _get_cluster_id_and_tasks_from_run_tasks(
     project_cluster_ids = defaultdict(list)
     all_cluster_tasks = defaultdict(list)
     for task in run["tasks"]:
-        if not exclude_tasks or task["task_key"] not in exclude_tasks:
+        if "cluster_instance" in task and (
+            not exclude_tasks or task["task_key"] not in exclude_tasks
+        ):
```
rmoneys, line 1038: Ignore tasks without a cluster too: https://sync-computing.slack.com/archives/C0405M188F4/p1695137323158079?thread_ts=1695136037.282939&cid=C0405M188F4

sync/_databricks.py
gorskysd: Ah have this fix in my branch too

gorskysd (approved):
LGTM

PR #45, opened 2023-10-12
[CUS-7] Add workspace API

rmoneys:
https://synccomputing.atlassian.net/browse/CUS-7

Add support for integrating workspaces and jobs to the developer API and CLI.


brandon-kaplan:
was able to use the cli end to end, have some 'user feedback' that could be part of a version 2.

the integrate workspace command needs a lot of flags and then extra input. I am not aware of what the best practices are but I feel like maybe from a user some of the flags could be moved to input?

Error handling, I think from the backend, that could be improved to be a little more descriptive:
`sync-cli workspaces apply-workspace-config 6142597627460088
Missing Permissions: You do not have permission to access this resource.`
Screenshot 2023-10-19 at 11 29 58 AM

Overall, that's nice to have stuff that shouldn't stop the progress of this PR and the PR works as is!

brandon-kaplan (approved):

rmoneys:
Thanks for the feedback! Here's my takeaway:

> the integrate workspace command needs a lot of flags and then extra input. I am not aware of what the best practices are but I feel like maybe from a user some of the flags could be moved to input?

Yeah, `create-workspace-config` input was separated by optionality and security - prompted input is both required and sensitive. Perhaps whenever prompt input is needed it's better to prompt for all input.

> Error handling, I think from the backend, that could be improved to be a little more descriptive:
sync-cli workspaces apply-workspace-config 6142597627460088 Missing Permissions: You do not have permission to access this resource.

I think that's an accurate message - not everyone in a tenant will have the permission to manage workspaces. The 500, however, error will be addressed in sync_backend.

PR #46, opened 2023-10-12
add support for dbfs cluster log locations

CaymanWilliams:
Adds handling for dbfs cluster log destinations so that the library can be used to set the cluster settings in Azure like it can with AWS. This is accompanied by some changes in backend to the projects table. The corresponding PR is: https://github.com/synccomputingcode/sync_backend/pull/414

I've tested backend/syncsparkpy locally with aws/azure through job_bank and it seems to be working


sync/_databricks.py
```diff
@@ -672,20 +762,32 @@ def get_project_cluster_settings(project_id: str, region_name: str = None) -> Re
...
+                            "destination": f"{cluster_log_dest.geturl()}/{project_id}",
                             "enable_encryption": True,
                             "region": region_name or boto.client("s3").meta.region_name,
                             "canned_acl": "bucket-owner-full-control",
                         }
                     }
                 }
             )
+
+        if cluster_log_dest.scheme == "dbfs":
```
rmoneys, line 780: `elif` is probably better here

sync/cli/_databricks.py
```diff
@@ -173,6 +174,19 @@ def create_submission(
     return
 
 
+@click.command
+@click.argument("project", callback=validate_project)
+def create_recommendation(project: dict):
+    return create_project_recommendation(project["id"])
```
rmoneys, line 180: What does it look like when you run this? It doesn't look like this command has any output.

sync/cli/_databricks.py
```diff
@@ -173,6 +174,19 @@ def create_submission(
...
+
+
+@click.command
+@click.argument("project", callback=validate_project)
+@click.argument("prediction-id")
+def get_recommendation(project: dict, prediction_id: str):
+    return get_project_recommendation(prediction_id, project["id"])
```
rmoneys, line 187: What does it look like when you run this? It doesn't look like this command has any output.

sync/cli/awsemr.py
```diff
@@ -117,6 +118,12 @@ def create_submission(run_id: str, project: dict, region: str = None):
     return
 
 
+@aws_emr.command
+@click.argument("project", callback=validate_project)
+def create_recommendation(project: dict):
+    return create_project_recommendation(project["id"])
```
rmoneys, line 124: Can we create a project recommendation for AWS EMR?

sync/cli/azuredatabricks.py
```diff
@@ -3,6 +3,7 @@
 from sync.cli._databricks import (
     access_report,
     create_prediction,
+    create_recommendation,
```
rmoneys, line 6: Do we want to add get_recommendation too?

sync/api/projects.py
```diff
@@ -269,3 +281,52 @@ def create_project_submission_with_eventlog_bytes(
         return Response(error=SubmissionError(message="Failed to upload event log"))
 
     return Response(result=response["result"]["submission_id"])
+
+
+def create_project_recommendation(project_id: str, **kwargs) -> Response[str]:
```
rmoneys, line 286: Can we be more specific for the developer API? There's only a couple valid options going in to kwargs here - let's call them out.

sync/clients/sync.py
```diff
@@ -119,6 +119,31 @@ def create_project_submission(self, project_id: str, submission: dict) -> dict:
             )
         )
 
+    def create_project_recommendation(self, project_id: str, **kwargs) -> dict:
```
rmoneys, line 122: `kwargs` is a very generic name. Maybe rec_params is better unless we want to be more explicit about the options for creating a rec - optoinal parmas in place of the catch-all ** param

sync/api/projects.py
rmoneys: If we don't want to advertise the options let's just go with `**options`

rmoneys (commented):
I have some questions about the CLI and project recommendations for aws-emr

sync/api/projects.py
chaneyw: Yea sorry, the idea was to not advertise the options. Happy to rename to **options.

sync/cli/awsemr.py
chaneyw: Good point, we cannot at the moment. I can delete this. I think there are changes to endpoints that would need to happen to support this for aws_emr (I think it just has to switch the task name that gets invoked for aws_emr) @gorskysd thoughts?

gorskysd: I think this has been taken care of already, but yes there's backend work that needs to be done to support emr.

sync/api/projects.py
```diff
@@ -269,3 +281,52 @@ def create_project_submission_with_eventlog_bytes(
...
+                return Response(result=result)
+            if result["state"] == "FAILURE":
+                return Response(error=RecommendationError(message="Recommendation failed"))
+        logger.info("Waiting for recommendation")
+        sleep(10)
+        response = get_project_recommendation(project_id, recommendation_id)
+
+
+def get_project_recommendation(project_id: str, recommendation_id: str) -> Response[dict]:
+    """ """
```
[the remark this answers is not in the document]
chaneyw: Added

sync/_databricks.py
chaneyw: Shouldn't ever be able to satisfy both conditions, but changed for safety sake

sync/cli/_databricks.py
chaneyw: Should now return a formatted JSON

sync/cli/awsemr.py
chaneyw: Deleted for now

sync/cli/azuredatabricks.py
chaneyw: Done

sync/clients/sync.py
```diff
@@ -171,8 +196,6 @@ def onboard_workflow(self, workspace_id, job_id: str, project_id: str) -> dict:
                 "POST",
                 "/integrations/v1/databricks/workflow/onboard",
                 params={"workspace_id": workspace_id},
-                headers=headers,
-                content=content,
```
rmoneys, line 175: Was this accidental?
https://api.synccomputing.com/docs#/Databricks%20Integration/onboard_workflow_integrations_v1_databricks_workflow_onboard_post

rmoneys (changes requested):

sync/clients/sync.py
CaymanWilliams: Yeah this was just a stupid merge issue

sync/api/projects.py
```diff
@@ -226,14 +238,34 @@ def create_project_submission(
     return Response(result=response["result"]["submission_id"])
 
 
+def clear_cluster_report_errors(cluster_report_orig: dict) -> dict:
```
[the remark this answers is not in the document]
CaymanWilliams: Wes wrote this one as a workaround for some issues we were having with certain Azure cluster reports where the error messages were messing up the client, preventing submissions. The end result was a `Too much data for declared Content-Length` error. I added a doc string, let me know if it needs to be formatted differently or needs more detail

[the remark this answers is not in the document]
rmoneys: Thanks. What comes after the colon is mostly free-form so the underscore isn't necessary, but formatting isn't so important since the leading underscore makes it no public and not on the doc page: https://synccomputingcode.github.io/syncsparkpy/

The explanation is very helpful.

rmoneys (approved):

PR #49, opened 2023-10-20
[PROD-1291] Apply prediction to existing job

rmoneys:
https://synccomputing.atlassian.net/browse/PROD-1291

Depends on https://github.com/synccomputingcode/syncsparkpy/pull/46


sync/_databricks.py
```diff
@@ -495,6 +499,82 @@ def record_run(
...
+            "job_clusters": [
+                {"job_cluster_key": project_cluster_path[1], "new_cluster": prediction_cluster}
+            ]
+        }
+    else:
+        new_settings = {
+            "tasks": [{"task_key": project_cluster_path[1], "new_cluster": prediction_cluster}]
+        }
+
+    response = databricks_client.update_job(job_id, new_settings)
```
singhals, line 570: My only comment on this is based on the databricks api documentation:

```Changes to the field JobSettings.timeout_seconds are applied to active runs. Changes to other fields are applied to future runs only.```


I don't believe we ever touch timeout_seconds but do we want to be extra cautious and delete it from `new_settings`? its a little bit silly but figured id call it out.

File: sync/_databricks.py [hunk 2]
```diff
@@ -582,6 +662,61 @@ def get_prediction_cluster(
...
+        }
+    else:
+        new_cluster_def = get_recommendation_cluster(
+            project_cluster_def, project_id, recommendation_id
+        )
+        new_settings = {
+            "tasks": [{"task_key": project_cluster_path[1], "new_cluster": new_cluster_def}]
+        }
+
+    response = databricks_client.update_job(job_id, new_settings)
```
singhals, line 712: Same comment here around making sure `timeout_seconds`. Can be deferred though.

sync/_databricks.py
```diff
@@ -582,6 +662,61 @@ def get_prediction_cluster(
...
+    job_clusters = _get_project_job_clusters(job)
+
+    project_cluster = job_clusters.get(project_id)
+    if not project_cluster:
+        if len(job_clusters) == 1 and None in job_clusters:
+            project_cluster = job_clusters[None]
+        else:
+            return Response(
+                error=DatabricksError(
+                    message=f"Unable to locate cluster in job {job_id} for project {project_id}"
```
singhals, line 689: Should we have a different error if job_clusters is empty? It seems like we return an empty signature if the project has more than one job cluster associated with it

singhals (approved):
I'm approving but it seems like "apply_project_recommendation" is not being exposed as a command. Was that intentional?

singhals:
Seems to be a Lint n' Test failure as well

sync/_databricks.py
rmoneys: Yeah I think that's a good idea here especially since the backend merges prior configs with predictions in the prediction API result - it could be set here. Less of a concern for recommendations but probably an easy fix there too.

rmoneys: Actually, we may be good here. The documentation only mentions `JobSettings.timeout_seconds` while we only touch clusters within `JobSettings`.
https://docs.databricks.com/api/workspace/jobs/update

sync/_databricks.py
rmoneys: > seems like we return an empty signature if the project has more than one job cluster associated with it

That's correct.

Yeah, I think that'd be good to be more explicit about each error scenario. I think this is good enough for now though - at least we have logging in `_get_project_job_clusters` for the scenario when there's more than 1 cluster for a project.

gorskysd (approved):
Looks good to me, very nice to additions to the library.

singhals (approved):

PR #50, opened 2023-10-22
combined submission and recommendation steps for GradientML

gorskysd:
Combine all the calls to submit data, create a rec, and get the `job settings` with the new rec applied all into one method. This saves the number of library calls required in a customers environment.

I thought it might be overkill to add this method to the CLI so I refrained from doing so for now.


sync/_databricks.py
```diff
@@ -296,9 +298,9 @@ def create_submission_for_run(
 
     if project_id in project_cluster_tasks:
         cluster_id, tasks = project_cluster_tasks.get(project_id)
-    elif None in project_cluster_tasks and len(project_cluster_tasks) == 1:
-        # If there's only 1 cluster and it's not tagged with project ID assume that's the one for the project
-        cluster_id, tasks = project_cluster_tasks.get(None)
+    elif len(project_cluster_tasks) == 1:
+        # If there's only 1 cluster then assume that's the one for the project
+        cluster_id, tasks = next(iter(project_cluster_tasks.values()))
```
gorskysd, line 303: I loosened this restriction because it was preventing me from "resubmitting" existing job runs to a new project-id. This is a super handy capability for development when we want to retest the submission/rec cycle with running new dbx jobs.

rmoneys: I'll allow this ;-)

sync/_databricks.py
```diff
@@ -672,6 +674,53 @@ def get_recommendation_cluster(
...
 
+def create_submission_and_get_recommendation_job_for_run(
+    job_id: str,
+    run_id: str,
+    plan_type: str,
+    compute_type: str,
+    project_id: str,
+    allow_incomplete_cluster_report: bool = False,
+    exclude_tasks: Union[Collection[str], None] = None,
+) -> Response[str]:
```
rmoneys, line 685: Looks like the result type is a dict

gorskysd:
@rmoneys Re: Use case. I was hoping to use this function in the bug bash tomorrow. You can see how I've applied it to that testing script here: https://github.com/synccomputingcode/sync_backend/blob/5af01fb799b730ad648abb2e8694969c80da5e16/gradient_test_scripts/gradient_full_test_aws_databricks.py#L141

This script follows the same prescription as in our "demo" notebook, where the library is used to get the Job Settings Update payload, and then the actual job update happens outside of the library.

gorskysd:
I should mention: it's ok if we don't want this functionality in `main` yet, especially with the "option to update job" functionality coming soon. I can always point people to this branch for the bug bash tomorrow.

PR #51, opened 2023-10-25
[PROD-1293] Add 'auto_appy_recs' property to projects

rmoneys:
https://synccomputing.atlassian.net/browse/PROD-1293


gorskysd (approved):
Also looks good. One step closer to take the wheel completely.

singhals (approved):

PR #52, opened 2023-10-25
Increase timeout period for get cluster state

aditya-jaishankar:
During gradient testing, it was noticed that there are cases where it takes longer than 300 seconds for a databricks cluster to spin down. This is incompatible with the default timeout period of 300 seconds and causes submission failures for successful job completions. The timeout is now increased to 600 seconds.

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -1057,7 +1057,7 @@ def terminate_cluster(cluster_id: str) -> Response[dict]:
 
 
 def _wait_for_cluster_termination(
-    cluster_id: str, timeout_seconds=300, poll_seconds=10
+    cluster_id: str, timeout_seconds=600, poll_seconds=10
 ) -> Response[dict]:
     start_seconds = time.time()
     cluster = get_default_client().get_cluster(cluster_id)
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -1059,6 +1059,7 @@ def terminate_cluster(cluster_id: str) -> Response[dict]:
 def _wait_for_cluster_termination(
     cluster_id: str, timeout_seconds=600, poll_seconds=10
 ) -> Response[dict]:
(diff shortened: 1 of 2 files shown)


sync/_databricks.py
```diff
@@ -1057,7 +1057,7 @@ def terminate_cluster(cluster_id: str) -> Response[dict]:
 
 
 def _wait_for_cluster_termination(
-    cluster_id: str, timeout_seconds=300, poll_seconds=10
+    cluster_id: str, timeout_seconds=600, poll_seconds=10
```
rmoneys, line 1060: Looks good. How about a logging statement just before the sleep though:
```
logging.info(f"Waiting for cluster {cluster_id} to terminate")
```

rmoneys (approved):

sync/_databricks.py
```diff
@@ -1057,8 +1057,9 @@ def terminate_cluster(cluster_id: str) -> Response[dict]:
...
 ) -> Response[dict]:
+    logging.info(f"Waiting for cluster {cluster_id} to terminate")
```
aditya-jaishankar, line 1062: @rmoneys Nice idea, made the change.

gorskysd (dismissed):
Small thing, but you might want to make sure this is updated to `main` and then increment the minor version in `sync/__init__.py`.

aditya-jaishankar:
> Small thing, but you might want to make sure this is updated to `main` and then increment the minor version in `sync/__init__.py`.

Thanks for pointing this out @gorskysd. Done!

gorskysd (approved):
LGTM

rmoneys:
Don't forget to always "Squash and merge"! Helps us make sense of the commits on main.

rmoneys:
And delete the PR branch after merge too - there should be a button to do that on this page.

aditya-jaishankar:
Whoops - will be better about this next time!

PR #53, opened 2023-11-02
[PROD-1282] Add workspace config attributes and update API paths

rmoneys:
https://synccomputing.atlassian.net/browse/PROD-1282

### DIFF
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.4.7"
+__version__ = "0.4.8"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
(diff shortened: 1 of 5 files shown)


brandon-kaplan (approved):
Approving because of the urgency, but as a follow up im wondering if we can get region from the unity catalog api's instead of getting it as input, since its static for the entire metastore

rmoneys:
@brandon-kaplan, can you point me to the API?

brandon-kaplan:
what if we used 
[Get metastore assignment for workspace](https://docs.databricks.com/api/workspace/metastores/current)
and then used that ID to [get metastore](https://docs.databricks.com/api/workspace/metastores/get) and used the region value from that? is that too many api calls and slowing things down?

rmoneys:
No - seems like a surer way to get the correct region

rmoneys:
FYI, started implementing this in the backend: https://github.com/synccomputingcode/sync_backend/pull/444/files#diff-afea6009de14705d2f7c045cedbde288336eeb6b118274adbf31462f52bf8a25

PR #54, opened 2023-11-06
add job id to project update

CaymanWilliams:
While doing some job_bank work, I realized that the project update method doesn't have a parameter for dbx job_id, even though the api allows it to be mutable (this could be useful in job_bank, but I could work around it too). If this is intentional, I'll just close this PR and delete the branch!


rmoneys (approved):
Sorry I overlooked this PR for so long. If the web API allows it then it's good to support it in the library, but I don't understand the use case. We wouldn't want to transfer a project to a different Databricks job, right?

Might be good to extend this to the CLI too: https://github.com/synccomputingcode/syncsparkpy/blob/fc5f69af0d334e6e3e697b00b2f7d9db21697435/sync/cli/projects.py#L118

gorskysd (approved):

brandon-kaplan (approved):

romainissynced (approved):
LGTM

CaymanWilliams:
Sorry all, needs another approval because of a conflict from Brandon's merge

brandon-kaplan (approved):

PR #59, opened 2023-11-13
PROD-1352 Add Profiles to CLI

brandon-kaplan:


brandon-kaplan:
Questions for scott -
1. I edited the config.py to make the profile switching changes. I dont _think_ it should affect the non cli files that import `API_KEY, CONFIG, DB_CONFIG`, like getting the `_clients`, but i wanted to know if im missing something in how those imports work and i need to update them as well.
2. I wasn't sure on how to add backward compatability so people on an existing cli version wouldnt have to make any changes, is that something i should try to figure out, or would communication to users be the better path
3. I wanted to confirm that these changes wouldnt break any existing implementations, like how the databricks workspace importing works as it is today.

brandon-kaplan:
I could also use some advice on why the automated tests are failing with `SyntaxError: annotated name '_active_profile' can't be global`
make test succeeds for me locally

sync/config.py
```diff
@@ -160,5 +201,47 @@ def __getattr__(name):
...
+
+
+def refresh_configurations(profile_name: str):
+    """Refreshes the global configurations to reflect the new active profile."""
+    global _active_profile, _api_key, _config, _db_config
+
+    _active_profile = profile_name
+    _api_key = None
+    _config = None
+    _db_config = None
```
rmoneys, line 218: Doesn't look like these `None` initializations are necessary

sync/config.py
```diff
@@ -141,12 +168,26 @@ def get_databricks_config() -> DatabricksConf:
...
 
 
+def get_profile_name() -> str:
+    """Gets the active profile
+
+    :return: active profile
+    :rtype: str
+    """
+    global _active_profile
+    _active_profile = _current_profile_dir().resolve().name
```
rmoneys, line 178: I'm not seeing where we depend on the gobal `_active_profile` being correctly set. I we don't depend on it then we're probably better off without it - won't have to worry about whether it correctly represents the state of the symlink.

sync/cli/profiles.py
```diff
@@ -0,0 +1,94 @@
...
+        click.echo("Available profiles:")
+        for profile in avail_profiles:
+            click.echo(f"  {profile}")
+
+
+@profiles.command()
+def active():
+    """Return the current profile."""
+    current_profile = get_profile_name()
+    if current_profile == "current":
```
rmoneys, line 88: When will this be true?

rmoneys (commented):
Also, I think it better to ditch the create profile command. Instead rely on `configure` with the `--profile` option. They kinda do the same thing anyway right? That'd allow us to move the `click` stuff out of `util` too - I'm not sure that's a good place to define the interface.

Profiles are kind of fundamental to everything else. Maybe it's better to remove the whole `profiles` command group - have them at the root instead, e.g.
```
sync-cli set-default-profile
sync-cli list-profiles
```

Have you thought about adding a `--profile` option at the command root kinda like how we do for `--debug`?

rmoneys:
Looks like this shouldn't effect modules where we import `CONFIG` and other settings. The thing to keep in mind there is that those configuration objects are static for the lifetime of the process - it doesn't look like this PR opens new doors to a dynamic configuration though.
A migration would be nice - would it be as easy as moving everything to a `default` profile dir if it doesn't exist? If we versioned the configuration that might make things easier for the next migration - wouldn't have to infer it then.
I don't think this breaks anything, but I can help run it through some core use-cases to be sure.

rmoneys:
Regarding the error in the automated tests, `make test` is failing locally for me. And the issue is easy to reproduce interactively:
```
>>> from sync.clients.sync import get_default_client                                                                                                                                                                                                                                                                                                                 
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                   
  File "<stdin>", line 1, in <module>                                                                                                                                                                                                                                                                                                                                
  File "/Users/scottromney/synchub/syncsparkpy/sync/clients/sync.py", line 6, in <module>                                                                                                                                                                                                                                                                            
    from ..config import API_KEY, CONFIG, APIKey                                                                                                                                                                                                                                                                                                                     
  File "/Users/scottromney/synchub/syncsparkpy/sync/config.py", line 190                                                                                                                                                                                                                                                                                             
    _active_profile: str = "default"                                                                                                                                                                                                                                                                                                                                 
    ^                                                                                                                                                                                                                                                                                                                                                                
SyntaxError: annotated name '_active_profile' can't be global
```
More on that here: https://stackoverflow.com/questions/67527942/why-cant-an-annotated-variable-be-global

The error still doesn't make complete sense to me since we are annotating it in the global scope, but maybe we can avoid it all together by getting rid of the global variable - https://github.com/synccomputingcode/syncsparkpy/pull/59#discussion_r1391559313

PR #60, opened 2023-11-15
add get-submission to the cli

gorskysd:
Mostly just to make our lives easier.

### DIFF
--- a/sync/clients/sync.py
+++ b/sync/clients/sync.py
@@ -144,6 +144,20 @@ class SyncClient(RetryableHTTPClient):
             )
         )
 
+    def get_project_submissions(self, project_id: str, params: dict = None) -> dict:
+        return self._send(
+            self._client.build_request(
+                "GET", f"/v1/projects/{project_id}/submissions", params=params
+            )
+        )
+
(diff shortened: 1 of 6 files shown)


sync/clients/sync.py
```diff
@@ -144,6 +144,20 @@ def get_project_recommendation(self, project_id: str, recommendation_id: str) ->
             )
         )
 
+    def get_project_submissions(self, project_id: str, params: dict = None) -> dict:
+        return self._send(
+            self._client.build_request(
+                "GET", f"/v1/projects/{project_id}/submissions", params=params
+            )
+        )
```
gorskysd, line 152: This one doesn't get used (yet), but I included for parity with the `get_project_recommendations` method elsewhere in this file.

sync/clients/sync.py
```diff
@@ -144,6 +144,20 @@ def get_project_recommendation(self, project_id: str, recommendation_id: str) ->
             )
         )
 
+    def get_project_submissions(self, project_id: str, params: dict = None) -> dict:
```
rmoneys, line 147: `**params` might make for a friendlier interface so long as we don't want to call any params out.

rmoneys (approved):

PR #63, opened 2023-11-28
make typing annotations compatible with >=3.7

CaymanWilliams:
Just ran into this while doing some testing. Some type annotations were incompatible with 3.7-3.10 versions of python.

### DIFF
--- a/sync/azuredatabricks.py
+++ b/sync/azuredatabricks.py
@@ -2,7 +2,7 @@ import logging
 import os
 import sys
 from time import sleep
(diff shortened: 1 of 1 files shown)


gorskysd (approved):
good catch! Might be worth taking a glance at the `awsdatabricks.py` file to be sure there aren't any slip ups there either

CaymanWilliams:
> good catch! Might be worth taking a glance at the `awsdatabricks.py` file to be sure there aren't any slip ups there either

Yeah, awsdatabricks imports fine!

PR #72, opened 2023-12-11
orjson replacement with json package

CaymanWilliams:
Passes all current tests, tested a few of the methods that changed individually (locally) but haven't done an exhaustive test of each method. It seems like everything should be working. The package differences are the options for encoding datetimes and the fact that json.dumps is a str and orjson.dumps is bytes. The echo calls in the cli don't care if the output is bytes or strings, and all other dumps calls have been cast to bytes. The utils folder has a custom class that encodes datetimes the same as orjson.

### DIFF
--- a/sync/azuredatabricks.py
+++ b/sync/azuredatabricks.py
@@ -59,6 +59,7 @@ from sync.models import (
     Response,
 )
 from sync.utils.dbfs import format_dbfs_filepath, write_dbfs_file
+from sync.utils.json import DefaultDateTimeEncoder
 
 __all__ = [
     "get_access_report",
@@ -374,7 +375,8 @@ def _monitor_cluster(
                         {
                             "instances": list(all_vms_by_id.values()),
                             "timelines": all_timelines,
-                        }
+                        },
+                        cls=DefaultDateTimeEncoder,
                     ),
                     "utf-8",
                 )
--- a/sync/azuredatabricks.py
+++ b/sync/azuredatabricks.py
@@ -5,7 +5,7 @@ from time import sleep
 from typing import List, Dict, Type, TypeVar, Optional
 from urllib.parse import urlparse
 
-import orjson
+import json
 from azure.common.credentials import get_cli_profile
 from azure.core.exceptions import ClientAuthenticationError
 from azure.identity import DefaultAzureCredential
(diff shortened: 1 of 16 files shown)


sync/azuredatabricks.py
```diff
@@ -371,11 +369,14 @@ def write_file(body: bytes):
...
-                        "timelines": all_timelines,
-                    }
+                bytes(
+                    json.dumps(
+                        {
+                            "instances": list(all_vms_by_id.values()),
+                            "timelines": all_timelines,
+                        }
+                    ),
+                    "utf-8",
```
gorskysd, line 379: I recall you saying that json dumps to a str while orjson dumps to bytes. Does the bytes format need to be adhered to here because the s3 write requires bytes? Just curious if/how this fails if you don't format it as bytes.

gorskysd (changes requested):
Looks like there are a few more orjson usages you missed.
image

Also, don't forget to bump the version!

CaymanWilliams:
> Looks like there are a few more orjson usages you missed. image
> 
> Also, don't forget to bump the version!

Oh man, looks like some files weren't saved but the changes were made locally so search isn't useful. Delightful. I'll go through and close all the tabs in my vscode and make sure they're saved

sync/azuredatabricks.py
CaymanWilliams: Yeah, if it's not cast into bytes you get a boto3 error that says something along the lines of file obj needs to be in bytes for s3. I don't recall the exact verbiage, but it was definitely a fatal error

CaymanWilliams: Well actually, that's true for this line https://github.com/synccomputingcode/syncsparkpy/blob/f97ebc090f91dfae4a73ec1aea9133e35212e2ea/sync/awsemr.py#L755. The write_file function you're talking about here expects bytes and manipulates the body expecting it to be bytes. It throws an error: `TypeError: a bytes-like object is required, not 'str'`. It looks like dbfs expects file content to be in b64.

gorskysd (dismissed):
Nice, everything looks good to me!

CaymanWilliams:
This last change fixes a bug where the _monitor_cluster methods were unable to write out the instance data because it needed it's own datetime encoder class in the json utils file. I've done some janky tests where I run diff on the new _monitor_cluster output and everything looks identical. The cli is unaffected, I've tested the clients and the encode_json method is working correctly for both clients.

gorskysd (approved):
Ran a successful test on both AWS and Azure. Looks good to me!

PR #76, opened 2023-12-20
Bugfix/prod 1477 dbfs filesize wait

gorskysd:
On Databricks Azure (possibly AWS too, but hasn't been observed), event log files being uploaded to DBFS may first appear with zero file size, then some time later become populated with data. Prior to this PR, the library would identify the zero-file-size files and upload them before they were actually populated with data.

This PR fixes the issue by doing to things:
1. Continues polling for new event log files while _any_ of the files have zero file size
2. Continues polling for new event log files if the data size changes between two consecutive polling periods

Jira Ticket: https://synccomputing.atlassian.net/browse/PROD-1477


Makefile
```diff
@@ -19,3 +19,9 @@ endif
 .PHONY: tidy
 tidy: format lint
 
+# Removes the directory that contains bytecode cache files
+# that are automatically generated by python.
+.PHONY: clean
+clean:
+	find . -type f -name "*.pyc" | xargs rm -fr
+	find . -type d -name __pycache__ | xargs rm -fr
```
gorskysd, line 27: This is convenience function for developers. Run `make clean` to delete pythons cache files.

brandon-kaplan (dismissed):

sync/_databricks.py
```diff
@@ -1725,6 +1725,25 @@ def _dbfs_directory_has_all_rollover_logs(contents: dict, run_end_time_millis: f
     )
 
 
+def _dbfs_any_file_has_zero_size(dbfs_contents: Dict):
```
aditya-jaishankar, line 1728: Super nitpicky but perhaps you also wanted a return type hint `-> bool | None`

aditya-jaishankar (dismissed):
Nice! Very clean way to implement the partial file checks.

sync/_databricks.py
gorskysd: good catch. Type hints are important, definitely not a nit :)

aditya-jaishankar (approved):
LGTM!

PR #82, opened 2024-01-02
Update Create Project

aditya-jaishankar:
For instance size optimization, the project models and database table now contain a new flag that holds the state of whether instance size optimization is enabled. This PR updates the library functions and endpoints that handle project creation and updates.

### DIFF
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.5.4"
+__version__ = "0.5.3"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.5.3"
+__version__ = "0.5.4"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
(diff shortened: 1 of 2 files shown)


sync/__init__.py
```diff
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "0.5.3"
```
aditya-jaishankar, line 2: Because this is part of a major new initiative (i.e. instance size optimization), does this deserve a more significant version number update. Should we make it 0.6.0? More generally, how should we be thinking about upgrades to version  number and what constitutes a major vs. a minor one, etc.?

syncbrian (commented):
Are we allowing the users to update this flag via the cli? At the moment, I only see an update to the api wit the new field being set to false and no updates to the actual cli command to allow users to flip the flag

syncbrian:
Cool makes sense, just wanted to verify that was the intention 👍

sync/__init__.py
syncbrian: Hmmm good question, I'm not sure what the versioning scheme has been in the past but agree that this is more than a simple fix and because it's not breaking changes the minor version update here would make sense. Unless @gorskysd or @singhals has some input here, I'd say we move this up to `0.6.0`

aditya-jaishankar: @syncbrian Based on our conversation from slack, I've changed the version back down to 0.5.3 (i.e. left it unchanged) with the thinking that the version number will be updated to 0.6.0 at the next release. Let me know if this sounds right to you

syncbrian: yup that sounds right!

syncbrian (approved):
LGTM!

PR #97, opened 2024-02-14
[PROD-1648] Bugfix describe volumes

aditya-jaishankar:
# Summary

There is a bug currently in the library where a call to the `ec2.describe_volumes` only accepts at most 199 items in the `Filters` argument. However, in the case of very large clusters, there can be more than 199 instance ids passed in, which causes the call to fail. 

This PR splits the number of filters to chunks of size no more than 199. 

# Testing Performed

* All unit tests pass
* Dummy projects were kicked off on databricks and the resulting cluster was inspected with the functions calls in both the cases of small number of workers as well as number of workers > 200 and it was verified that the functions calls return the cluster volume data as expected. 

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-1648)

### DIFF
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "1.0.0"
+__version__ = "1.0.1"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/awsdatabricks.py
+++ b/sync/awsdatabricks.py
@@ -2,7 +2,7 @@ import json
 import logging
 from pathlib import Path
 from time import sleep
-from typing import List, Tuple
+from typing import Iterator, List, Tuple
 from urllib.parse import urlparse
 
 import boto3 as boto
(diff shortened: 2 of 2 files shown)


sync/awsdatabricks.py
```diff
@@ -481,28 +481,39 @@ def _get_ebs_volumes_for_instances(
 ) -> List[dict]:
     """Get all ebs volumes associated with a list of instance reservations"""
 
+    def get_chunk(instance_ids: list, chunk_size: int) -> Iterator[list]:
```
romainissynced, line 484: Isn't the return type technically a `-> Generator[list, None, None]`?

Might want to also type the args like so:

```suggestion
    def get_chunk(instance_ids: List[str], chunk_size: int) -> Generator[List[str], None, None]:
```

romainissynced (approved):
Looks good, just left a nitpick on typing that doesn't need to be fixed before merge.

sync/awsdatabricks.py
romainissynced: Mentioning this because it can help the IDE auto-complete.

For example if you were to `[item.count('a') for item in self.get_chunk(***)` pycharm will autocomplete the `.count()` because it knows its a list of strings

sync/__init__.py
```diff
@@ -1,4 +1,4 @@
 """Library for leveraging the power of Sync"""
-__version__ = "1.0.0"
```
gorskysd, line 2: Need to update this to 1.0.3 now

sync/awsdatabricks.py
```diff
@@ -481,28 +481,39 @@ def _get_ebs_volumes_for_instances(
...
+
+            response = ec2_client.describe_volumes(Filters=filters)
+            volumes = response.get("Volumes", [])
             next_token = response.get("NextToken")
 
+            while next_token:
+                response = ec2_client.describe_volumes(Filters=filters, NextToken=next_token)
+                volumes += response.get("Volumes", [])
+                next_token = response.get("NextToken")
+
```
gorskysd, line 516: Have you had a chance to test this out on a big cluster yet?

aditya-jaishankar: Yes, I tested this out by kicking off a big job with 300 .large instances and then wrote a local scipt to call `describe_volumes()` on that cluster_id while the cluster was running and it worked fine (also worked fine for a cluster with 10 instances)

PR #99, opened 2024-02-15
[PRO-16170] Give ability to set keys, create submission with cluster report, and …

singhals:
# Summary

* Expose get_cluster_activity_events as a public function
* Add mechanism to stop monitor cluster polling when we are hosting it until the cluster has reached a termination state
* Create submission with all the cluster report info coming in from outside the library

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-) (add id)

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -219,7 +219,8 @@ def _get_run_information(
     cluster_report = cluster_report_response.result
     if cluster_report:
         cluster = cluster_report.cluster
-        eventlog = _get_event_log_from_cluster(cluster, tasks)
+        eventlog_response = _get_event_log_from_cluster(cluster, tasks)
+        eventlog = eventlog_response.result
         if eventlog:
             return Response(result=(cluster_report, eventlog))
 
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -219,12 +219,14 @@ def _get_run_information(
     cluster_report = cluster_report_response.result
     if cluster_report:
         cluster = cluster_report.cluster
-        return _get_event_log_from_cluster(cluster, tasks)
+        eventlog = _get_event_log_from_cluster(cluster, tasks)
+        if eventlog:
+            return Response(result=(cluster_report, eventlog))
 
     return cluster_report_response
 
 
-def _get_event_log_from_cluster(cluster: Dict, tasks: List[Dict]) -> Response[str]:
+def _get_event_log_from_cluster(cluster: Dict, tasks: List[Dict]) -> Response[bytes]:
     spark_context_id = _get_run_spark_context_id(tasks)
     end_time = max(task["end_time"] for task in tasks)
     eventlog_response = _get_eventlog(cluster, spark_context_id.result, end_time)
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -18,7 +18,14 @@ import boto3 as boto
 from sync.api import projects
 from sync.clients.databricks import get_default_client
 from sync.config import CONFIG  # noqa F401
-from sync.models import DatabricksAPIError, DatabricksClusterReport, DatabricksError, Response
+from sync.models import (
+    DatabricksAPIError,
+    DatabricksClusterReport,
+    DatabricksError,
+    DatabricksComputeType,
+    DatabricksPlanType,
+    Response
+)
 from sync.utils.dbfs import format_dbfs_filepath, read_dbfs_file
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


sync/_databricks.py
```diff
@@ -56,6 +63,58 @@ def get_cluster(cluster_id: str) -> Response[dict]:
...
+        return project_response
+    cluster_path = project_response.result.get("cluster_path")
+
+    project_cluster_tasks = _get_project_cluster_tasks(run, project_id, cluster_path)
+
+    cluster_tasks = project_cluster_tasks.get(project_id)
+    if not cluster_tasks:
+        return Response(
+            error=DatabricksError(
+                message=f"Failed to locate cluster in run {run_id} for project {project_id}"
```
taylorgaw, line 93: To further improve the "debuggability" of this message, do we want to include cluster path or some info about the the cluster tasks?

singhals:
> At first pass, the logic here is 👍🏻. My concern is that a lot of the method parameters have `Dict` types, which isn't great for future code readability. I don't think this is a "must have", but it would be nice to have these dicts defined as types of their own so that other developers know how to interact with these objects

Agreed but we were trying to avoid making a drastic changes here while dealing with some legacy code that is used in the library today.

sync/_databricks.py
singhals: Ya let me revisit this when we add azure support. Its a good point!

taylorgaw (dismissed):

sync/_databricks.py
```diff
@@ -240,6 +307,17 @@ def _get_cluster_report(
     raise NotImplementedError()
 
 
+def _create_cluster_report(
+        cluster: dict,
+        cluster_info: dict,
+        cluster_activity_events: dict,
+        tasks: List[dict],
```
gorskysd, line 314: I think these type hints need to be `Dict` to maintain the older python compatibility.

sync/awsdatabricks.py
```diff
@@ -412,6 +443,11 @@ def _monitor_cluster(
                     "utf-8",
                 )
             )
+
+            if kill_on_termination:
+                cluster_state = get_default_client().get_cluster(cluster_id).get("state")
+                if cluster_state == "TERMINATED":
+                    while_condition = False
```
gorskysd, line 450: Do we actually need the `kill_on_termination` flag at all? At a glance, it looks like if you did away with it but kept the check for TERMINATED then we're set for hosted monitoring and the existing functionality is unchanged (existing functionality being the monitoring running on the job cluster, in which case it's killed when the cluster terminates anyway).

sync/config.py
```diff
@@ -138,6 +146,14 @@ def get_databricks_config() -> DatabricksConf:
     return _db_config
 
 
+def set_databricks_config(db_config: DatabricksConf):
+    global _db_config
+    if _db_config is not None:
+        raise RuntimeError("Databricks config has already been set and the library does not support resetting "
+                           "credentials")
+    _db_config = db_config
```
gorskysd, line 154: This is cool, nice.

sync/azuredatabricks.py
```diff
@@ -223,6 +228,25 @@ def _get_cluster_report(
     )
 
 
+def _create_cluster_report(
+        cluster: dict,
+        cluster_info: dict,
+        cluster_activity_events: dict,
+        tasks: List[dict],
```
gorskysd, line 235: `Dict` type hints here again.

gorskysd (changes requested):
There are a few `dict` type hints that I think need to get changed to `Dict` to maintain compatibility prior to python 3.9. Also, might be able to clean up the monitoring flag, but not blocking on this. Otherwise looks good -- congratulations on navigating the hot mess of the library ;)

sync/_databricks.py
singhals: Interesting. Theres a nice mix between the two already in the library. ill go with Dict though

sync/awsdatabricks.py
singhals: This is mostly a safeguard on changing behavior in existing functionality and not needing to check for the cluster state repeatedly from the init script. When we use hosted monitoring, we set kill_on_termination to `true`

sync/_databricks.py
gorskysd: Oh hm, maybe it's only an issue when you try to type the contents of the dict. E.g. I'm almost certain that `dict[str]` would fail ... but maybe just a play `dict` is fine.

gorskysd (approved):
Looks like the `Dict` change is unnecessary afterall, and it already exists elsewhere in the library. LGTM!

PR #103, opened 2024-03-04
feat: get diff between latest recommendation and cluster schema [PROD-1686]

eheinlein-sync:
# Summary

We're adding functionality to display cluster recommendations, as well as a side by side comparison and a merged cluster definition.

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-1686) (PROD-1686)


sync/clients/sync.py
```diff
@@ -194,6 +235,32 @@ def onboard_workflow(self, workspace_id, job_id: str, project_id: str) -> dict:
             )
         )
 
+    def get_latest_project_config_recommendation(self, project_id: str) -> ProjectConfiguration:
```
romainissynced, line 238: Are you sure this returns `ProjectConfiguration`, it looks like its returning a `dict` below

sync/clients/sync.py
```diff
@@ -194,6 +235,32 @@ def onboard_workflow(self, workspace_id, job_id: str, project_id: str) -> dict:
...
+        latest_recommendation = self.get_latest_project_recommendation(project_id)
+        if latest_recommendation.get("result"):
+            return latest_recommendation["result"][0]["recommendation"]["configuration"]
+        else:
+            return {"error": f"No project recommendation found for Project {project_id}"}
+
+    def get_cluster_definition_and_recommendation(
+        self, project_id: str, cluster_spec_str: str
```
romainissynced, line 246: Why is `cluster_spec_str` here and in `get_updated_cluster_defintion` a string and not a dict?

Seems weird to me to load json instead of passing a dict directly.

sync/clients/sync.py
```diff
@@ -194,6 +235,32 @@ def onboard_workflow(self, workspace_id, job_id: str, project_id: str) -> dict:
...
+    ) -> dict:
+        cluster_recommendation = self.get_latest_project_config_recommendation(project_id)
+        return {
+            "cluster_recommendations": cluster_recommendation,
+            "cluster_definition": json.loads(cluster_spec_str),
+        }
+
+    def get_updated_cluster_defintion(
```
romainissynced, line 254: Does this perform the `diff`?

sync/clients/sync.py
```diff
@@ -1,14 +1,48 @@
...
+class DBFSClusterLogConfiguration(BaseModel):
+    destination: str
+
+
+class ClusterLogConfiguration(BaseModel):
+    s3: S3ClusterLogConfiguration
+    dbfs: DBFSClusterLogConfiguration
+
+
+class ProjectConfiguration(BaseModel):
```
singhals, line 31: We should move this to aws databricks and add one for azure databricks as well

sync/clients/sync.py
```diff
@@ -1,14 +1,48 @@
...
+    node_type_id: str
+    driver_node_type: str
+    custom_tags: Dict
+    cluster_log_conf: ClusterLogConfiguration
```
singhals, line 35: Should this be a Union[S3ClusterLogConfiguration, DBFSClusterLogConfiguration] ? it can be either or but can't be both or neither

singhals (changes requested):

sync/clients/sync.py
taylorgaw: This produces an updated cluster definition with the recommendations merged in. The "diff" is in the method above which displays both the original cluster definition as well as the recommendations. Kartik and I decided it's best to provide the user with all of the information they need, and to do the diff-ing on their end

sync/clients/sync.py
taylorgaw: I made this parameter a string because my understanding is that the cluster spec would be loaded in the command line as text

sync/clients/sync.py
taylorgaw: Good catch. I'll think of a better way to handle the case if there is no Recommendation provided

sync/clients/sync.py
```diff
@@ -122,6 +156,13 @@ def get_project_recommendation(self, project_id: str, recommendation_id: str) ->
             )
         )
 
+    def get_latest_project_recommendation(self, project_id: str) -> dict:
```
syncbrian, line 159: we should know what this request response object looks like right? May be worth updating this `dict` to a project recommendation schema

taylorgaw: For the API calls, I've been following the pattern of the other calls, and that's why it's a `dict`. I agree we should have a schema for this, but it might make more sense to have a large overhaul of the API response types

sync/api/projects.py
```diff
@@ -396,3 +405,79 @@ def get_project_submission(project_id: str, submission_id: str) -> Response[dict
...
+    :rtype: dict
+    """
+    recommendation_response = get_latest_project_config_recommendation(project_id)
+    if not recommendation_response:
+        logger.info(f"No cluster recommendation found for {project_id}")
+        return Response(error=RecommendationError(message="Recommendation failed"))
+    response_str = json.dumps(recommendation_response.result)
+    return Response(
+        result={
+            "cluster_recommendations": json.loads(response_str),
```
singhals, line 447: Its one recommendation, no? "cluster_recommendations" -> "cluster_recommendation"?

sync/api/projects.py
```diff
@@ -396,3 +405,79 @@ def get_project_submission(project_id: str, submission_id: str) -> Response[dict
...
+            "cluster_definition": json.loads(cluster_spec_str),
+        }
+    )
+
+
+def get_updated_cluster_defintion(
+    project_id: str, cluster_spec_str: str
+) -> Response[Union[AWSProjectConfiguration, AzureProjectConfiguration]]:
+    """Print Cluster Definition merged with Project Configuration Recommendations.
```
singhals, line 456: I think you meant "Return" and not "Print" here

sync/api/projects.py
```diff
@@ -396,3 +405,79 @@ def get_project_submission(project_id: str, submission_id: str) -> Response[dict
...
+        latest_recommendation = json.loads(latest_rec_str)
+        cluster_definition = json.loads(cluster_spec_str)
+        for key in latest_recommendation.keys():
+            cluster_definition[key] = latest_recommendation[key]
+
+        # instance_source and driver_instance_source are not
+        # included in recommendation and need to be updated as well
+        driver_recommendation = cluster_definition["node_type_id"]
+        cluster_definition["instance_source"] = {"node_type_id": driver_recommendation}
+        cluster_definition["driver_instance_source"] = {"node_type_id": driver_recommendation}
```
singhals, line 479: We should do a deep merge here like we do when we apply a recommendation from the library:
https://github.com/synccomputingcode/syncsparkpy/blob/dc7a217d4131fcd91ec1f9d3982d9e5c9ed1cbeb/sync/_databricks.py#L633

singhals: it might be good to review this pathway before merging this:

https://github.com/synccomputingcode/syncsparkpy/blob/dc7a217d4131fcd91ec1f9d3982d9e5c9ed1cbeb/sync/_databricks.py#L524

I think we are rewriting logic that exists already

singhals (changes requested):

Makefile
```diff
@@ -3,7 +3,7 @@ FILES := $(shell git diff --name-only --diff-filter=AM $$(git merge-base origin/
 
 .PHONY: test
 test:
-	pytest
+	pytest -vv
```
taylorgaw, line 6: This is consistent across our other repos for `make test`

singhals (approved):

PR #110, opened 2024-04-11
Prod-1820 forma add sync library function to fetch cluster spec of latest submission in project

brandon-kaplan:
# Summary

*please add a few lines to give the reviewer context on the changes*
Adds the ability to get the latest submission's configuration for a project, and an option that specifies whether it needs to be the last successful submission, or the most recent regardless of failure. Also adds this to the CLI. Added a separate file for projects as this doesnt really fall under azure/aws databricks, and added tests as well

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-1820)

Screenshot 2024-04-11 at 3 52 54 PM
Screenshot 2024-04-11 at 3 53 47 PM

### DIFF
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,5 +1,5 @@
 """Library for leveraging the power of Sync"""
 
-__version__ = "1.4.2"
+__version__ = "1.5.0"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/__init__.py
+++ b/sync/__init__.py
@@ -1,5 +1,5 @@
 """Library for leveraging the power of Sync"""
 
-__version__ = "1.4.1"
+__version__ = "1.4.2"
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
--- a/sync/api/projects.py
+++ b/sync/api/projects.py
@@ -1,5 +1,6 @@
 """Project functions
 """
+
 import io
 import json
 import logging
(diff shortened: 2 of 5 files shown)


sync/projects.py
```diff
@@ -0,0 +1,51 @@
...
+
+class NoSubmissionsFoundError(SubmissionRetrievalError):
+    """No submissions were found."""
+
+
+class NoSuccessfulSubmissionsFoundError(SubmissionRetrievalError):
+    """No successful submissions were found."""
+
+
+def get_latest_submission_config(project_id: str, success_only: bool = False) -> dict:
```
singhals, line 16: I think we want success by default

sync/projects.py
```diff
@@ -0,0 +1,51 @@
...
+            f"Failed to retrieve submissions for project '{project_id}'. {e}"
+        ) from e
+
+    if not submissions:
+        raise NoSubmissionsFoundError(f"No submissions found for project '{project_id}'.")
+
+    if success_only:
+        latest_successful_submission = next(
+            (submission for submission in submissions if submission["state"] == "SUCCESS"), None
+        )
```
singhals, line 40: what about ordering by date?

sync/api/projects.py
```diff
@@ -416,6 +417,19 @@ def get_project_submission(project_id: str, submission_id: str) -> Response[dict
...
 
+def get_submissions(project_id: str) -> Response[dict]:
+    """Get submissions for a project id
+
+    :param project_id: project ID
+    :type project_id: str
+    :return: List of Submission Configuration objects
+    :rtype: dict
+    """
+    recent_submissions = get_default_client().get_project_submissions(project_id)
```
singhals, line 428: might consider page=0&per_page=1 if we support it. that should get you the latest? not sure

singhals (changes requested):

sync/projects.py
brandon-kaplan: i wasnt sure on that to be honest. That seems like a preference thing right? Like latest inherently means the latest without conditions, no? Happy to change it though

sync/projects.py
brandon-kaplan: The result from the db is already returned ordered by date descending. even though it's a list it would retain that order

singhals: ya my worry is that its a weak api contract. for now its fine but we have to remember that the sort order never changes.

brandon-kaplan: yeah that is true

singhals (dismissed):

sync/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Library for leveraging the power of Sync"""
 
-__version__ = "1.4.1"
+__version__ = "1.4.2"
```
singhals, line 3: Should be a minor update, no? this is not a bug fix.

singhals (approved):

PR #111, opened 2024-04-17
Prod 1735 update cli to allow for hosted workspace creation

brandon-kaplan:
# Summary

*please add a few lines to give the reviewer context on the changes*
This updates the CLI for Creating and Updating a workspace. The cli was a bit outdated in this regard. I restructured the azure/aws fields to be dependent on compute provider, which is either implicitly discovered via the url, or prompted if they have a custom domain. 
Azure Client Id, Client Secret, Tenant Id, and Subscription Id were added to an azure hosted creation.
aws iam role arn was added for aws hosted creation.

For updating, we were already following a pattern of 'only update provided values' so i kept that going for this. Updated values will continue to be passed in as explicit flags to get updated. I intentionally left out the ability to switch between compute providers and to switch between hosting types. The former because it makes more sense for a new workspace (and really unlikely) and the latter because we need to include updating existing workflows with the proper webhook and that felt out of scope for this.

Forma Bug Fix: While working on this forma reached out for a version bump on boto: it's included in this PR
## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-1735)

**Creating an Azure Workspace:**
Screenshot 2024-04-17 at 4 24 52 PM

**Updating a field in the azure workspace**
Screenshot 2024-04-17 at 4 25 57 PM

**Creating an AWS Workspace**
Screenshot 2024-04-17 at 4 53 24 PM

**Updating an AWS Workspace**
Screenshot 2024-04-17 at 4 56 49 PM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,44 +11,44 @@ build-backend = "hatchling.build"
 
 [project]
 name = "syncsparkpy"
-authors = [{"name" = "Sync Computing"}]
+authors = [{ "name" = "Sync Computing" }]
 readme = "README.md"
 requires-python = ">=3.7"
 classifiers = [
-  "Intended Audience :: Information Technology",
-  "Intended Audience :: System Administrators",
-  "Operating System :: OS Independent",
-  "Development Status :: 3 - Alpha",
-  "License :: OSI Approved :: Apache Software License",
-  "Programming Language :: Python",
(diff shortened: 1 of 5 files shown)


syncbrian (approved):
Nice udpates and refactoring! LGTM

PR #121, opened 2024-06-03
[PROD-2015] Add save_cluster_report and refactor _monitor_cluster

romainissynced:
# Summary

Eventbridge collector [PR](https://github.com/synccomputingcode/sync-svc-library-tasks/pull/54) access to save_cluster_report without calling _monitor_cluster.

I extracted the save cluster report logic into its own function. The collect cluster report task can now call this new function, the old _monitor_cluster function is now also using the new function.

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-) (add id)


sync/awsdatabricks.py
```diff
@@ -350,6 +351,61 @@ def _get_aws_cluster_info_from_s3(bucket: str, file_key: str, cluster_id):
...
+def save_cluster_report(
+    cluster_id: str,
+    instance_timelines: list[dict],
+    cluster_log_destination: tuple[str, ...] | None = None,
+    cluster_report_destination_override: dict[str, str] | None = None,
+) -> bool:
+    cluster = get_default_client().get_cluster(cluster_id)
+    spark_context_id = cluster.get("spark_context_id")
+
+    if not spark_context_id:
```
romainissynced, line 363: @brandon-kaplan just to double check, spark_context_id only exists when the cluster is running right?

I may want to not check for eventbridge cause we could process events after the cluster is done running.

Do you see any caveats with saving the cluster report after the cluster is done running? Below `_get_ec2_instances` fetches the current running instances, I'll take a look on Monday.

brandon-kaplan: it gets a value when the cluster starts running but it doesn't disappear after its completed, ill ping you directly since I'm replying so late

brandon-kaplan (dismissed):
I don't see anything obviously wrong with what you did. I think the conflicts are due to this branch missing the 'monitor once' methods which is what the svc layer is currently using for hosted. So I think the risk is low to move forward with testing

brandon-kaplan (approved):

PR #126, opened 2024-07-22
fix cluster report creation bug

CaymanWilliams:
# Summary

Occasionally, when the dbx api would fail for miscellaneous reasons, we would enter a section of code on cluster creation that was causing an exception. I've fixed the exception but would like to note that this code path will no longer yield a submission because we'll be missing an instance timeline and the other sync-specific information. It's no longer sufficient for us to grab a best guess snapshot. Should we keep this code in here so we at least get all available data in the cluster report saved to s3 or should we take these code paths out for aws/azure because this cluster report is deprecated?

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-2005)

### DIFF
--- a/sync/azuredatabricks.py
+++ b/sync/azuredatabricks.py
@@ -284,34 +284,17 @@ def _get_cluster_instances(cluster: dict) -> Response[dict]:
             json.loads(cluster_instances_file_response) if cluster_instances_file_response else None
         )
 
-    # If this cluster does not have the "Sync agent" configured, attempt a best-effort snapshot of the instances that
-    #  are associated with this cluster
-    if not cluster_instances:
-        resource_group_name = _get_databricks_resource_group_name()
+        if not cluster_instances:
+            no_instances_message = (
+                f"Unable to find cluster information from Sync's cluster monitor for cluster `{cluster_id}`. "
+            )
+            return Response(error=DatabricksError(message=no_instances_message))
 
-        compute = _get_azure_client(ComputeManagementClient)
-        if resource_group_name:
-            vms = compute.virtual_machines.list(resource_group_name=resource_group_name)
-        else:
-            logger.warning("Failed to find Databricks managed resource group")
-            vms = compute.virtual_machines.list_all()
-
-        cluster_instances = {
-            "instances": [
-                vm.as_dict()
-                for vm in vms
-                if vm.tags.get("Vendor") == "Databricks"
-                and vm.tags.get("ClusterId") == cluster["cluster_id"]
-            ]
-        }
-
-    if not cluster_instances.get('instances', None):
-        no_instances_message = (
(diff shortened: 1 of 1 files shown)


sync/azuredatabricks.py
```diff
@@ -292,11 +292,6 @@ def _get_cluster_instances(cluster: dict) -> Response[dict]:
         compute = _get_azure_client(ComputeManagementClient)
         if resource_group_name:
             vms = compute.virtual_machines.list(resource_group_name=resource_group_name)
-
-            for vm in vms:
-                compute.virtual_machines.instance_view(
-                    resource_group_name=resource_group_name,
-                )
```
CaymanWilliams, line 299: Don't know why this was here in the first place, but the vms object can only be iterated through once so it even after fixing the `instance_view` call it would cause cluster instances to be empty

shughes-uk (approved):

romainissynced (approved):
LGTM

As for the question, I would remove deprecated code

sync/azuredatabricks.py
```diff
@@ -284,39 +284,17 @@ def _get_cluster_instances(cluster: dict) -> Response[dict]:
...
-        no_instances_message = (
-            f"Unable to find any active or recently terminated instances for cluster `{cluster_id}`. "
-            + "Please refer to the following documentation for options on how to address this - "
-            + "https://docs.synccomputing.com/sync-gradient/integrating-with-gradient/databricks-workflows"
+    else:
+        no_cluster_logs_message = (
+            f"No cluster log configuration has been set for cluster `{cluster_id}`. "
         )
-        return Response(error=DatabricksError(message=no_instances_message))
+        return Response(error=DatabricksError(message=no_cluster_logs_message))
```
gorskysd, line 297: This last `no cluster log location` seems like pretty substantial change  to how we handle no cluster log being set. I do think there's some improvement to be made in our handling, but I'm concerned that putting it in this PR might affect our monitoring in unexpected ways. Also, if we start rejecting submissions more eagerly we won't get cluster reports we otherwise could.

This might be a good parking lot topic. Also @romainissynced or @singhals might have their own opinions.

singhals: We probably dont want to do this. This will prevent creating a submission altogether. The preference here will be to show this as a "Pending Setup: Cluster logs location not set" on the frontend of the projects page and the project details page. That way we can still assert cost but also confirm that we have an e2e submission processing even if its without the event log. We can do this by using the databricks_job table and @taylorgaw is working on modeling the cluster configs into the databricks_cluster table where we can check if the cluster log location is set or not

gorskysd (changes requested):
Remove gating on cluster log location. Otherwise the cleanup is good.

gorskysd (approved):
LGTM

PR #132, opened 2024-08-05
[PROD-2276] Add new class to cached auth token to file (or redis with subclass)

romainissynced:
# Summary

- Changed the auth flow so that you can pass a CachedToken cls
- Made new class to cache token in file cache
- Made sure that the file cache works with the lowest py version that we support, py 3.7 

## Checklist

[Related Jira Ticket](https://synccomputing.atlassian.net/browse/PROD-2276)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,8 @@ dependencies = [
     "azure-mgmt-compute==30.1.0",
     "azure-mgmt-resource==23.0.1",
     "azure-cli-core==2.50.0",
-    "platformdirs"
+    "platformdirs",
+    "python-dateutil==2.9",
 ]
 dynamic = ["version", "description"]
 
@@ -50,7 +51,6 @@ dev = [
     "Sphinx==4.3.0",
     "respx==0.20.1",
     "deepdiff==6.3.0",
-    "python-dateutil==2.9",
 ]
 
 [tool.hatch.version]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ dev = [
     "Sphinx==4.3.0",
     "respx==0.20.1",
     "deepdiff==6.3.0",
+    "python-dateutil==2.9",
 ]
 
 [tool.hatch.version]
--- a/sync/clients/sync.py
+++ b/sync/clients/sync.py
@@ -2,8 +2,9 @@ import json
 import logging
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
-from typing import Generator, Optional, Tuple
+from typing import Generator, Optional, Tuple, Type
 
+import dateutil.parser
 import httpx
 from platformdirs import user_cache_dir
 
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -50,6 +50,7 @@ dev = [
     "Sphinx==4.3.0",
     "respx==0.20.1",
     "deepdiff==6.3.0",
+    "python-dateutil==2.9",
```
romainissynced, line 53: this was already installed from botocore, pinning the version here to be safe

sync/clients/sync.py
```diff
@@ -13,78 +14,110 @@
...
-        if self._access_token_expires_at_utc:
-            return datetime.now(
-                tz=timezone.utc
-            ) < self._access_token_expires_at_utc - timedelta(seconds=20)
-        return False
+
+class SyncAuth(httpx.Auth):
+    requires_response_body = True
+
+    def __init__(self, api_url: str, api_key: APIKey, cached_token_cls: Type[CachedToken] = FileCachedToken):
```
romainissynced, line 99: The plan here is to subclass `CachedToken` from `sync-svc-library-tasks` to support redis caching, the file cache behavior should match what was there before.

sync/clients/sync.py
```diff
@@ -94,11 +127,9 @@ def build_auth_request(self) -> httpx.Request:
     def update_access_token(self, response: httpx.Response):
         if response.status_code == httpx.codes.OK:
             auth = response.json()
-            self._access_token = auth["result"]["access_token"]
-            self._access_token_expires_at_utc = auth["result"]["expires_at_utc"]
-            self._set_cached_token(
-                self._access_token, self._access_token_expires_at_utc
-            )
+            access_token = auth["result"]["access_token"]
+            expires_at_utc = dateutil.parser.isoparse(auth["result"]["expires_at_utc"])
```
romainissynced, line 131: Had to do this because py3.7 can't parse a datetime with timezone with builtin `fromisoformat`

shughes-uk: boo!

shughes-uk (dismissed):

pyproject.toml
```diff
@@ -35,7 +35,8 @@ dependencies = [
     "azure-mgmt-compute==30.1.0",
     "azure-mgmt-resource==23.0.1",
     "azure-cli-core==2.50.0",
-    "platformdirs"
+    "platformdirs",
+    "python-dateutil==2.9",
```
shughes-uk, line 39: I suspect this pin will cause some pain fyi

singhals: Is this internal pain or for our customers?

shughes-uk: our customers, strict pins like this are somewhat painful.

singhals: Ya agreed - lets figure out a good plan after this pr gets out to address the pinned versions.

singhals (approved):

pyproject.toml
romainissynced: Made a ticket to track that: https://synccomputing.atlassian.net/browse/PROD-2324

Can bring it up at parking lot, would pinning as <3 be safe as that's the next major version that might introduce breaks?

PR #136, opened 2024-08-13
EBF: Enable azure retries 

vinooganesh:
The while loop for azure eventlog retrieval is never entered because eventlog_response is populated after the first call.

### DIFF
--- a/sync/_databricks.py
+++ b/sync/_databricks.py
@@ -1481,6 +1481,7 @@ def _poll_for_eventlog_from_dbfs(
                 f"No or incomplete event log data detected - attempting again in {poll_duration_seconds} seconds"
             )
             sleep(poll_duration_seconds)
+            eventlog_response = None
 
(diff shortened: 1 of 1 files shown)


sync/_databricks.py
```diff
@@ -1481,6 +1481,7 @@ def _poll_for_eventlog_from_dbfs(
                 f"No or incomplete event log data detected - attempting again in {poll_duration_seconds} seconds"
             )
             sleep(poll_duration_seconds)
+            eventlog_response = None
```
shughes-uk, line 1484: the s3 fetcher above this has the exact same issue btw

shughes-uk (dismissed):

sync/_databricks.py
vinooganesh: Updated!

shughes-uk (approved):

